PR #1, opened 2015-06-03
launch modal with player

tyard:


css/bcgs_ford_main.css
```diff
@@ -0,0 +1,25 @@
...
+
+#modal-container {
+    background-color: #fff;
+    margin: 80px auto 0px auto;
+    width: 800px;
+    height: 500px;
+}
+
+#modal-player .video-js {
+    width: 100%;
```
bceacuna: shouldn't we prefix all our stuff with bcc-?

lib/bcgs_ford_main.js
```diff
@@ -0,0 +1,87 @@
...
+    initial drop
+
+*/
+(function(window, document) {
+
+    "use strict";
+    
+    // replace name of plugin, then proceed to constructor function
+    var NAME = "FordPlayer",
+
```
bceacuna: are we going to use our new naming convention on this project? bccFordPlayer.

lib/bcc_core.js
```diff
@@ -0,0 +1,378 @@
+/**
+
+Copyright(c) 2014 Brightcove, Inc.
+
```
bceacuna: We need to update our template. Copyright(c) 2015 Brightcove, Inc.

lib/bcc_core.js
```diff
@@ -0,0 +1,378 @@
...
+                    find = function(selector) {
+                        return document.querySelector(selector);
+                    },
+
+                    /**
+                     * Creates an event dispatching model of set properties
+                     * with explicit getters and setters.
+                     */
+                    Model = function(p) {
+                        var properties = {};
```
bceacuna: Can you provide an example of how to use this method?

css/bcgs_ford_main.css
tyard: It is only necessary for our reusable components when we don't know the projects they will be used in. For a customer-specific project it isn't necessary, especially in a case like this.

lib/bcgs_ford_main.js
tyard: I think only necessary for reusable components. bcc shouldn't be a prefix for client-specific code.

lib/bcc_core.js
tyard: Sure, I will update the comment. But in the meantime:
```
var model = new bcc.Model({
    state:0, view:null
});
model.bind(bcc.events.CHANGE, function(event) {
  console.log( model.getState() );
}
model.setState(1);

PR #2, opened 2015-06-09
Extend modal

bceacuna:
Implemented FORD-7, FORD-8, and FORD-9

Styles and multi-device support will be implemented next sprint.


css/bcgs_ford_main.css
```diff
@@ -20,9 +23,24 @@
...
+}
+
+#modal-container.active {
+    /*TODO: update this after getting the final designs*/
+    height: 600px;
+    -webkit-transition: height 300ms ease-in-out;
+    -moz-transition: height 300ms ease-in-out;
+    -o-transition: height 300ms ease-in-out;
+    transition: height 300ms ease-in-out;
 }
```
tyard: You only need the transitions in one place (the base selector) not here as well.

css/bcgs_ford_main.css
```diff
@@ -20,9 +23,24 @@
...
 
 #modal-player .video-js {
-    width: 100%;
-    height: 450px;
+   height: 450px;
+    width: 100%
```
tyard: Not sure why the swap, but make sure the indentation is correct.

css/bcgs_ford_modal.css
```diff
@@ -0,0 +1,107 @@
...
+	padding-bottom: 57%; /*TODO: update with the aspect ratio of the content*/
+	height: 0;
+}
+
+.bcc-modal-close-button {
+    cursor:pointer;
+    text-align: center;
+    vertical-align: middle;
+    line-height: 1.2em;        /* the same as your div height */
+    visibility: visible;
```
tyard: Why the need to set this?

css/bcgs_ford_modal.css
```diff
@@ -0,0 +1,107 @@
...
+    font-size: 20px;
+    position: absolute;
+    padding: 10px 10px 10px 10px;
+    width: 1.2em; 
+    height: 1.2em; 
+    bottom: auto;
+    top: 0px;
+    left: auto;
```
tyard: I don't think you need this if you are already setting right to 0px.

css/bcgs_ford_modal.css
```diff
@@ -0,0 +1,107 @@
...
+    right: 0px;
+    opacity: 0.85;
+    background: #121111;
+    color: white;
+
+}
+
+.vjs-user-inactive .bcc-modal-close-button {
+    -webkit-transition-delay: 3s; /* Safari */
+    transition-delay: 3s;
```
tyard: I would suggest putting your transitions in the base selector.

css/bcgs_ford_modal.css
```diff
@@ -0,0 +1,107 @@
...
+.bcc-detail-center {
+    display: inline-block; 
+    width: 15%;       
+}
+
+/**
+TODO: we might need to set the content in js to support localization.
+*/
+.bcc-detail-right:before{
+        content: "Show Details \25bc";
```
tyard: Yeah, we will need it elsewhere.

Can you point me to the doc or discussion where you found that before content should always be used? I can't say I've ever used that (or had a need to that I didn't know) so I'd like to understand why it is done here.

css/bcgs_ford_modal.css
```diff
@@ -0,0 +1,107 @@
...
+        cursor:pointer;
+        float: right; 
+        width: 15%; 
+        font-size: 12px;
+        color: cornflowerblue;
```
tyard: placeholder color?

css/bcgs_ford_modal.css
```diff
@@ -0,0 +1,107 @@
...
+        left:0px;
+        position: absolute; }
+  TODO: ask for the width breakpoint for mobile.
+*/        
+@media only screen and (max-width: 568px) {
+    #modal-container {
+        margin: 0px;
+        width: 100%;
+        height: 100%;       
+    }
```
tyard: We shouldn't have selectors from the main file overridden here. This should all either be in the main file, or the main file should control everything for modal-container. Having definitions in two different files will become confusing.

index.html
```diff
@@ -31,17 +32,21 @@
...
             <div id="modal-container">
-                 <video
-                    id="player"
-                    data-account="2254903859001"
-                    data-player="da67ceac-64d4-46aa-bff3-e98a97da0edf"
-                    data-embed="default"
-                    class="video-js" controls>
-                </video>
-                <script src="//players.brightcove.net/2254903859001/da67ceac-64d4-46aa-bff3-e98a97da0edf_default/index.min.js"></script>
+                <div class="bcc-wrapper-intrinsic-ratio">
```
tyard: I'd rather wrap this using JS, as opposed to requiring Team Detroit to have to have a div within a div within a div with very specific IDs. We can get away with modal-player and modal-container, but for something so under-the-hood as bcc-wrapper-intrinsic-ratio I'd rather this be hidden from them in our scripts.

lib/bcgs_ford_modal.js
```diff
@@ -0,0 +1,147 @@
...
+                            
+         
+        /**
+        * Handler: Shows/hides video detail panel
+        */
+        onShowHideDetails = function(event){
+            log("Show/hide details");
+            //Resizes modal window and shows details
+            $(".bcc-detail-right").toggleClass("active");
+            $("#modal-container").toggleClass("active");
```
tyard: As opposed to having two classes it is better practice to have a single class at the top level that all nested elements can listen to (like the player does). So I would suggest that you just add the class "active" (or probably something more descriptive like "details-opened") to modal-container, then all divs nested there can respond to that single class (so instead of ".bcc-detail-right.active" you'd have ".details-opened .bcc-detail-right").

lib/bcgs_ford_modal.js
```diff
@@ -0,0 +1,147 @@
...
+            log(video.description);
+        };
+
+    if (player.mediainfo) {
+         setVideoInfo();
+    } else {
+        player.on("loadstart", setVideoInfo);
+    }
+            
+    player.ready(function(){
```
tyard: Have you found the need to do all of this on ready? I haven't seen adding listeners or divs to the player when the plugin is initialized as problematic. I'd like to understand the cases when this is necessary.

css/bcgs_ford_main.css
bceacuna: I'll fix this. I might have swapped this when I made the changes to implement Desktop first.

css/bcgs_ford_modal.css
bceacuna: good catch. We don't need to set visibility here.

css/bcgs_ford_modal.css
bceacuna: You are right. I'll remove right and bottom.

css/bcgs_ford_modal.css
bceacuna: This just hide the close button immediately after 3 seconds. I have not implemented any transition effect to the button,

Which is the base selector for you?
As for the transition effect, fade in/out is ok?

css/bcgs_ford_modal.css
bceacuna: Since we need to support localization, I made some changes in a new branch and removed  the pseudo class :before. Now I am setting the content from javascript.

https://developer.mozilla.org/en-US/docs/Web/CSS/content

If you don't use :before with the property "content", the content is not displayed in firefox (this is not an issue in Chrome)

css/bcgs_ford_modal.css
bceacuna: This is just a random color that I picked from my IDE color component. I'll apply the styles this week.

css/bcgs_ford_modal.css
bceacuna: I just realized that one of my TODO: comments was stuck in my home computer and was never pushed to this branch. 

/**
* TODO: if there is no objection, I'll move all the styles from bcgs_ford_main.css to bcgs_ford_modal.js next sprint. 
* I just want to keep everything related to the modal window in one place.
* 
*/

index.html
bceacuna: No problem, I'll make the necessary changes on the current sprint

lib/bcgs_ford_modal.js
bceacuna: Thanks for the recommendation, I was thinking about changing this today.

lib/bcgs_ford_modal.js
bceacuna: To tell you the true, I took this from the plugin example in our documentation. I am not sure yet when this is necessary. I'll remove it to make the code cleaner.

bceacuna:
Todd. 
I have already started working on the new sprint and made some changes to the styles. so if there is no objection, I will take all your suggestions,  implement them to the current sprint, and then delete this PR.

Let me know if that's ok with you

Thanks

css/bcgs_ford_modal.css
tyard: I would put it on .bcc-modal-close-buton itself.

css/bcgs_ford_modal.css
tyard: I guess my question was why use "content" in general?

tyard:
Sounds good.

css/bcgs_ford_modal.css
bceacuna: I believe it is a best practice to change the content  directly in css when the element change states. I found this trick online.

tyard: If you can find a link, that would be great."Best Practice" vs. "CSS Trick" would be a good distinction to understand -- what exactly are the benefits? I couldn't find anything myself on a quick search on why you would want to do this as a best practice for an element changing state. As long as we know why we are doing it, I have no problem with using the method. I just want to understand it, and make sure we all know why.

index.html
bceacuna: What would be the best way to wrap this using JS? I was trying to do it using jquery.wrapall() but I couldn't make it work.  

<div class="bcc-wrapper-intrinsic-ratio">
</video>
<script src="//players.brightcove.net/2254903859001/da67ceac-64d4-46aa-bff3-e98a97da0edf_default/index.min.js"></script>
</div>

tyard: I don't know why wrap wouldn't work -- I'd have to try it and let you know -- there's no reason it shouldn't.

Failing that, we don't actually need this div. You can get the same sizing with some simple JS alone, which is what we've done in the past. But there's no reason that wrap, in some form, shouldn't work.

PR #3, opened 2015-06-11
Style detail panel Desktop

bceacuna:
This is a work in  progress. 

*Known issues*
1. Styles have not been fully implemented in Desktop
2. Video detail panel is not displayed in mobile devices
3. The modal window does not fit in small screens (i.e laptops). You will need to zoom out your browse to at lease 60% to see the detail panel

The implementation works fine on Firefox, Chorme and Safari (Mac)


css/bcgs_ford_main.css
```diff
@@ -15,14 +28,164 @@
...
+
+.bcc-padding-left {
+    padding-left: 60px;
+}
+
+.bcc-save-icon {   
+    position:absolute;
+    top: 0px;
+    right: 60px;       
+}
```
tyard: Things should only be added as the sprint dictates. I see stuff in here that we aren't delivering (save button, build button), yet at the same time we are talking about not having enough time during the sprint. It creates more work for you, heavier reviews, and unexpected deliveries to QA. Please try to contain work to what was agreed upon.

css/bcgs_ford_main.css
```diff
@@ -15,14 +28,164 @@
...
+
+.bcc-detail-right{
+    font-size: 16px;
+    cursor:pointer;
+    float: right; 
+    width: 15%; 
+    height: 96px;
+    font-size: 12px;
+    color: cornflowerblue;
```
tyard: make sure placeholders have a @todo

css/bcgs_ford_main.css
```diff
@@ -15,14 +28,164 @@
...
+.bcc-more-detail-panel{
+    position: relative;
+    top: 20px;
+}
+
+.bcc-detail-description{
+    width: 70%;
+}
+
+.bcc-build-button {
```
tyard: CTAs will not always be "build" -- please remove this for now as it is not part of this sprint and shouldn't be included in this file. In addition, I think we need to discuss implementation prior to this commit since CTAs should be handled more generically.

lib/bcgs_ford_modal.js
```diff
@@ -0,0 +1,161 @@
...
+    log(NAME + " loaded.");
+    // TODO: rename css class names (something more descriptives)
+    var video,
+        $closeButtonDiv = $('<div class="bcc-modal-close-button">x</div>'),
+        $mobileTemplate = $('<div class="bcc-detail-panel">  \
+                                <div>1-5</div> \
+                                <div>Save Icon</div> \
+                                <div>Video Title</div> \
+                                <div>Details</div> \
+                                <div >Build & Price </div> \
```
tyard: This will not be how we implement these as they will be configurable and localized. These shouldn't have been added in this sprint or PR. Can you please remove to keep this consistent with the sprint task? Thank you.

lib/bcgs_ford_modal.js
```diff
@@ -0,0 +1,161 @@
...
+                            <div class="bcc-detail-left bcc-align-detail-tab-vertically"></div> \
+                            <div class="bcc-detail-center bcc-align-detail-tab-vertically">1 of 5</div> \
+                            <div class="bcc-detail-right bcc-align-detail-tab-vertically">Show Details</div> \
+                            <hr/>\
+                            <div class="bcc-more-detail-panel"> \
+                                <div class="bcc-detail-description bcc-padding-left"></div> \
+                                <div class="bcc-build-button">Build & Price </div> \
+                                <div class="bcc-save-icon"> Save Icon </div> \
+                                <div class="bcc-detail-scrollable-panel bcc-padding-left"> Scrollable Area </div> \
+                            </div> \
```
tyard: Same here.

lib/bcgs_ford_modal.js
```diff
@@ -0,0 +1,161 @@
...
+                        </div>'), 
+                             
+        // Since the layouts for Desktop and Mobile are different, we use two different templates
+        // to build the video detail panel.
+        //TODO: implement.isMobile.Phone (only devices with small screens)
+        $detailDiv = bcc.isMobile.iPhone() ? $mobileTemplate : $desktopTemplate,
```
tyard: I'm not sure I agree with two different templates. Let's discuss. We should be able to control, hide and relayout out pretty much anything with CSS.

lib/bcgs_ford_modal.js
```diff
@@ -0,0 +1,161 @@
...
+             
+        /**
+        * Sets the video object and updates Video details panel.
+        */
+        setVideoInfo = function(event) {
+           log("----- set video info");
+            video = player.mediainfo;
+            $detailDiv.find('.bcc-detail-left').html(video.name);
+            $detailDiv.find('.bcc-detail-description').html(video.long_description);
+            $detailDiv.find('.bcc-detail-scrollable-panel').html("voluptua eu. Sea ut molestie recteque, Vel id accusam propriae adversarium. Ei agam melius his. Sed odio vivendum laboramus ne. Oportere similique duo ne, ex mea putent explicari. Mel brute maiorum detracto id, nullam convenire vituperata no his. Ex populo nemore sit, menandri instructior qui et. voluptua eu. Sea ut molestie recteque, Vel id accusam propriae adversarium. Ei agam melius his. Sed odio vivendum laboramus ne. Oportere similique duo ne, ex mea putent explicari. Mel brute maiorum detracto id, nullam convenire vituperata no his. Ex populo nemore sit, menandri instructior qui et.");
```
tyard: What is this? The details at this time should only have title and description.

lib/bcgs_ford_modal.js
```diff
@@ -0,0 +1,161 @@
...
+            log(video.description);
```
tyard: Is this log needed? Doesn't seem helpful in the commit.

tyard:
I think the things that are not in the sprint should be removed for now, and we should talk about the two template approach -- I still think we could achieve the same through CSS and avoid some confusion -- but in general should be fine to merge after some tweaks.

One thing we need to solve, though, is that this review took twice as long as it should because so much of it was listed here as new even though it had already been reviewed in the previous PR. So either we need to resolve previous PRs first before moving on to new ones that present much of the same code marked as new, or this PR should be against the previous PR's branch (is this dif'd against dev instead), or find some other solution to enable us to see JUST what's changed. It will speed up turnaround of these code reviews.

css/bcgs_ford_main.css
bceacuna: Got it.

lib/bcgs_ford_modal.js
bceacuna: Yes, let's discuss this next time you are in.

bceacuna:
I just removed all the things that are not part of the sprint made all the changes that you suggested. Please review the code one more time and if everything is fine merge away and push everything to our test environment (http://psdev.brightcove.com/ford)

As for the review, my bad, this PR should have been against the previous PR's branch. I'll do that next time I decide not to merge a PR.

PR #4, opened 2015-06-12
use autostart instead of play

tyard:


bceacuna:
everything looks fine.

PR #5, opened 2015-06-12
style tweaks, resize player

tyard:


css/bcgs_ford_main.css
```diff
@@ -3,12 +3,12 @@
...
-	src: url('fonts/regular/extralight.eot?#iefix') format('embedded-opentype'),
-	     url('fonts/regular/extralight.woff') format('woff'),
-	     url('fonts/regular/extralight.ttf') format('truetype'),
-	     url('fonts/regular/extralight.svg#webfont') format('svg');
+    font-family: 'antennaExtraLight';
+    src: url('fonts/regular/extralight.eot');
+    src: url('fonts/regular/extralight.eot?#iefix') format('embedded-opentype'),
+         url('fonts/regular/extralight.woff') format('woff'),
+         url('fonts/regular/extralight.ttf') format('truetype'),
+         url('fonts/regular/extralight.svg#webfont') format('svg');
```
tyard: Just indentation change -- replaced tabs.

css/fonts/regular/_notes/dwsync.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
tyard: hmm, added this to ignore -- not sure why it got pushed.

bceacuna:
Looks good.

lib/bcgs_ford_modal.js
```diff
@@ -98,6 +103,31 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
+        sizeModal = function() {
+            var aspectRatio = 0.5624, /* 703/1250 */
+                windowHeight = $(window).height(),
+                maxWidth = $(window).width()-margin*2,
+                maxHeight = windowHeight-margin*2,
+                minHeight = maxWidth*aspectRatio+openDrawerHeight,
+                modalWidth = minHeight > maxHeight ? (maxHeight-openDrawerHeight)/aspectRatio : maxWidth,
+                videoHeight = modalWidth*aspectRatio,
+                modalHeight = videoHeight;
+
```
bceacuna: I would suggest expressing the aspectRatio as a fraction (16/9).

lib/bcgs_ford_modal.js
```diff
@@ -43,15 +43,18 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
     log(NAME + " loaded.");
     // TODO: rename css class names (something more descriptives)
     var video,
+        $modal = $("#modal-container"),
+        margin = 50, /* @todo: get this from CSS */
+        drawerHeight = 100,
+        openDrawerHeight = 250,
         $closeButtonDiv = $('<div class="bcc-modal-close-button">x</div>'),
```
bceacuna: The close button icon has been replaced with an image. So the character "x" should be removed

lib/bcgs_ford_modal.js
tyard: I was taking from the CSS that was there, which 1250/703, which is not quite 16/9. We can certainly use that (it's cleaner), but I didn't know if that matched exactly the aspect ratio you had discovered and put in the CSS. I'm for updating it, though.

lib/bcgs_ford_modal.js
tyard: Yep, missed that.

css/bcgs_ford_main.css
```diff
@@ -41,32 +39,25 @@
...
-    height: 1.2em; 
+    width: 66px; 
+    height: 65px; 
     top: 0px;
     right: 0px;
-    opacity: 0.85;
-    background: #121111;
-    color: white;
+    background-image: url(../img/close-icon.png);
 }
```
bceacuna: Shouldn't be using an svg instead, so the image is properly rendered/resized on devices with a high PPI.

tyard: Perhaps, but I've never seen much issue with it, and at this time I was just rushing to get something out to Alfredo that looked like the designs. We can revisit as we test on different sizes. What was there before wasn't SVG so this change wasn't to remove it anyway.

PR #9, opened 2015-06-16
add playlist API

tyard:
bcc.model has been created, which can be subscribed to listen for model changes. Properties include video and playlist.

bcc.model.getPlaylist() is itself a model, which can be subscribed to hear about when its selected index changes. Properties include videos and selectedIndex.

To play the next item in the playlist:

bcc.trigger(bcc.events.REQUEST_NEXT_VIDEO);

To play the previous item in the playlist:

bcc.trigger(bcc.events.REQUEST_PREVIOUS_VIDEO);

(I realize I need a way to play a specific index. That can be done with bcc.model.getPlaylist().setSelectedIndex(), but I'd rather tie this to an event so the plugins are not directly coupled. I'll put that into the next PR.)


lib/bcgs_ford_main.js
```diff
@@ -49,25 +49,82 @@ v1.0.0: 6/2/2015
...
+            bcc.unbind(bcc.events.REQUEST_NEXT_VIDEO);
+            bcc.unbind(bcc.events.REQUEST_PREVIOUS_VIDEO);
+            
+            if (selectedIndex >= config.videos.length) selectedIndex = 0; 
+            // create new playlist model from config video
+            playlist = new bcc.Model({
+                selectedIndex: selectedIndex,
+                videos: config.videos
+            });
+            // bind to request for previous/next
```
bceacuna: Can you please explain what's the purpose of calling bcc.model.getPlaylist() above (line 60) ? It seems that playlist is not used until it is set again here.

bceacuna:
Looks good.

lib/bcgs_ford_main.js
tyard: Good catch. At one time I was using it to unbind its listener. I removed that line, but not this one. I think I will actually add that line back in.

PR #10, opened 2015-06-17
event of setting playlist index

tyard:


lib/bcgs_ford_main.js
```diff
@@ -50,21 +50,22 @@ v1.0.0: 6/2/2015
...
+            var playlist = bcc.model.getPlaylist() || new bcc.EventDispatcher(),
                 selectedIndex = config.playIndex ? config.playIndex : 0;
             
             bcc.ModalWindow.open();
             
             // clear previous handlers
-            bcc.unbind(bcc.events.REQUEST_NEXT_VIDEO);
-            bcc.unbind(bcc.events.REQUEST_PREVIOUS_VIDEO);
+            bcc.unbind();
+            playlist.unbind();
```
bceacuna: I assume you are removing all the handlers here. if so,  I would suggest updating  bcc_core.unbind() API doc and mention the fact that calling the method without parameters remove all the added handlers.

lib/bcgs_ford_main.js
```diff
@@ -85,6 +86,12 @@ v1.0.0: 6/2/2015
                     playlist.setSelectedIndex(index-1);
                 }
             });
+            bcc.bind(bcc.events.REQUEST_VIDEO_INDEX, function(event) {
+                var index = event.data;
+                if (index > -1 && index < playlist.getVideos().length) {
+                    playlist.setSelectedIndex(index);
+                }
+            });
             // add playlist to main model
```
bceacuna: Can we use bcc.unbind() to remove handlers that were added as anonymous functions?

lib/bcgs_ford_main.js
tyard: sure

lib/bcgs_ford_main.js
tyard: You can't remove a specific handler without a reference. But you can remove all handlers for an event ( bcc.unbind(type) ) or all handlers for all events ( bcc.unbind() ).

bceacuna:
Looks good.

PR #11, opened 2015-06-17
implemented save button

bceacuna:
This implementation only supports desktop


index.html
```diff
@@ -45,7 +45,14 @@
                 <script src="lib/bcc_web.js"></script>
                 <script src="lib/bcgs_ford_main.js"></script>
                  <script src="lib/bcgs_ford_modal.js"></script>
-                <script>videojs("player").FordPlayer({debug:true});</script>
+                <script>videojs("player").FordPlayer({
+                        debug:true,                   
+                        showSaveButton: true,
+                        callbackSave: function(video){ 
+                            console.log("save video: " + video.id);
```
tyard: Should be videoId.

lib/bcgs_ford_modal.js
```diff
@@ -152,6 +153,16 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
     $detailDiv.find('.bcc-detail-right').html("Show Details &or;");
     $detailDiv.find('.bcc-detail-right').click(onShowHideDetails);
+            
+    //Shows Save button based on player configuration
+    if (bcc.settings.showSaveButton){
+        var $saveButton = $detailDiv.find('.bcc-save-icon');
+        $saveButton.show();   
+        // Adds event handler
+        $saveButton.on("click touchstart", function(){ 
+            bcc.settings.callbackSave(video)
```
tyard: semicolon

tyard: Actually, you want to pass the videoConfig, not the BC video object.

index.html
tyard: Captured as a later comment, but it's videoId because this should be the video config, not the video.

bceacuna: Ok, I thought we had to pass the video object. Let me fix this.

lib/bcgs_ford_modal.js
bceacuna: I just pushed my changes. Let me know if it is ok to merge.

lib/bcgs_ford_modal.js
```diff
@@ -152,6 +153,16 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
     $detailDiv.find('.bcc-detail-right').html("Show Details &or;");
     $detailDiv.find('.bcc-detail-right').click(onShowHideDetails);
+            
+    //Shows Save button based on player configuration
+    if (bcc.settings.showSaveButton){
+        var $saveButton = $detailDiv.find('.bcc-save-icon');
+        $saveButton.show();   
+        // Adds event handler
+        $saveButton.on("click touchstart", function(){ 
+            bcc.settings.callbackSave(bcc.model.getVideo());
```
tyard: This is already stored in videoConfig. This is fine, but it's another call that isn't strictly necessary.

lib/bcgs_ford_modal.js
```diff
@@ -152,6 +153,16 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
     //TODO: implement localization
     $detailDiv.find('.bcc-detail-right').html("Show Details &or;");
     $detailDiv.find('.bcc-detail-right').click(onShowHideDetails);
+            
+    //Shows Save button based on player configuration
+    if (bcc.settings.showSaveButton){
+        var $saveButton = $detailDiv.find('.bcc-save-icon');
```
tyard: Just something to consider going forward -- defining a local variable this far down in the code can cause potential problems (like hoisting) as the code grows in complexity).

Generally, we try to declare all vars at the top of the scope to make it clear what is defined in that scope, even if the assignments don't come till later.

tyard:
ok to merge, or after the changes suggested

lib/bcgs_ford_modal.js
bceacuna: hoisting... good catch. I'll fix this

lib/bcgs_ford_modal.js
bceacuna: I replaced bcc.model.getVideo() with videoConfig

PR #13, opened 2015-06-18
Committing carousel files into repository

BC-Mikey:


tyard:
In general, the text "catalog" was specific to the Samsung app "Catalog" -- should probably make it generic for a reusable carousel component.

index.html
```diff
@@ -14,14 +14,54 @@
             }
         </style>
         <link rel="stylesheet" type="text/css" href="css/bcgs_ford_main.css" ref=stylesheet>
+        <link rel="stylesheet" type="text/css" href="css/bcgs_carousel.css" ref=stylesheet>
```
tyard: I want to keep the reusable stuff as "bcc_" for now -- we would do a full repo renaming at one point, if we want. Better than having some bcc_ and some bcgs_ (client stuff we will continue with bcgs_, it's just the reusable library I'm talking about).

index.html
```diff
@@ -40,12 +80,29 @@
...
                 <script src="lib/bcgs_ford_main.js"></script>
-                 <script src="lib/bcgs_ford_modal.js"></script>
-                <script>videojs("player").FordPlayer({debug:true});</script>
+                <script src="lib/bcgs_ford_modal.js"></script>
+                
+                <!-- carousel scripts -->
+                <script src="lib/bcc_carousel.js"></script>
+                <script src="lib/catalogItems.js"></script>
+                <script src="lib/navigationButton.js"></script>
+
```
tyard: If it's client specific, let's use bcgs_{client}_component. If it's reusable, let's use bcc_{component}.

For now at least.

lib/bcc_carousel.js
```diff
@@ -0,0 +1,638 @@
...
+                log = bcc.log,
+                
+                // flag for debug mode
+                debug = bcc.debug;
+            
+/* ============================================================================================ */
+/* ALL CODE SHOULD BE PLACED WITHIN FOLLOWING FUNCTION                                          */
+/* ============================================================================================ */
+
+    var Carousel,
```
tyard: I'm assuming most of this (except the boilerplate) is as it was with the Met plus the stuff we code reviewed elsewhere, right (the swiping)?

lib/bcgs_ford_main.js
```diff
@@ -144,7 +170,7 @@ v1.0.0: 6/2/2015
 
     // if this is for a Brightcove Player, add bcc as a VideoJS plugin
     if (window.videojs) {
-        videojs.plugin(NAME, constructor);
+        var play = videojs.plugin(NAME, constructor);
```
tyard: What is this?

lib/bcgs_ford_main.js
```diff
@@ -67,6 +76,23 @@ v1.0.0: 6/2/2015
...
+                false
+            );
+            
+            $carousel = $(".catalog-carousel");
+            
+            if (!bcc.isMobile.any()) {
+                $nextBtn = new bcc.NavigationButton('next-btn', $carousel);
+                $prevBtn = new bcc.NavigationButton('prev-btn', $carousel);
+            }
+
```
tyard: I think we want to keep the carousel logic out of main (just as Modal is a separate file). This shouldn't have UI stuff in it, I don't think.

index.html
```diff
@@ -40,12 +80,29 @@
                     data-embed="default"
                     class="video-js" controls autoplay>
                 </video>
+                <!-- carousel content will be placed in below div -->
+                <div id="bcc-carousel-content"></div>
```
tyard: We don't want to force Team Detroit to include a bunch of HTMl on their page for the player -- this should be done within our JS. Please refer to the docs I've already provided them (look in ActiveCollab) about the player set up.

bcgs_ford_filmstrip.js should be made it's own plugin that adds UI to the modal and instantiates the carousel. Let's keep all the UI in that separate file.

lib/catalogItems.js
```diff
@@ -0,0 +1,100 @@
...
+    -pared down HTML to only Ford-specific elements
+    -using 'click' event to avoid confusion with carousel 'touchend' event handler
+
+*/
+(function(window, document) {
+
+    "use strict";
+    
+    // replace name of plugin, then proceed to constructor function
+    var NAME = "CatalogItem",
```
tyard: Again, let's make this generic and reusable -- perhaps bccCarouselItem?

tyard: Actually, I wonder about the reusability of this anyway. Really, the CatalogItem in the Samsung app was just an HTML blob, and this will change in pretty much any implementation we do. We could keep this class, and then allow an HTML blob to be passed in, but basically that's the entirety of this class, with a lot of useless boilerplate then.

I think what we should do is lose this class, then in the new bcgs_ford_filmstrip define the carousel item HTML that will be passed to the carousel.

lib/navigationButton.js
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Navigates items to the left.
+     */
+    NavigationButton.prototype.navigateLeft = function(event) {
+        if (this.index-this.totalRows < 0) { return false; }
+        this.index -= this.totalRows;
+        this.highlightIndex(this.index);
+        this.animate(1);
+        return true;
+    };
```
tyard: I'm not sure I am understanding this class. What is "this" in this scenario? I am assuming the carousel, but how is that set up that something assigned to the NavigationButton.prototype is referring to an external object as this?

Also, you have code in here referring to indices and rows and functions like animate that don't exist here. This is tightly coupling this class with some external class that isn't evident from this code.

Unless I'm completely misreading this we should discuss the best way to handle going forward. I think we just need generic buttons (which don't need a separate class/file) that call functions already defined in the carousel.

lib/bcc_carousel.js
BC-Mikey: Yup, haven't changed anything within bcc_carousel besides the boilerplate code and adding this to the beginning to avoid JS errors:
var Carousel,
        WrappingCarousel,
        GridCarousel,

All other methods and functional code has remained the same.

lib/bcgs_ford_main.js
BC-Mikey: Woops, reverting.

index.html
BC-Mikey: Ah thanks! I was going to ask about this in our stand-up.

lib/navigationButton.js
BC-Mikey: You're reading it right. I don't actually use anything in this class except for:

```javascript
var NavigationButton = function(buttonClass, $parent) {
        var $navBtn = $("<div class='" + buttonClass + "' />");
        
        // append to parent class
        $parent.append($navBtn);
};
```

The rest should actually be deleted as its leftover from when I first started building the class. I'm now wondering if this is too small for its own component class?

PR #14, opened 2015-06-18
adding filmstrip counter

tyard:


lib/bcgs_ford_modal.js
```diff
@@ -122,21 +126,62 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
             $modal.removeClass("no-transition");
         },
         
         /**
+         * Sizes title based on other elements.
+         */
+        sizeTitle = function() {
+            if ($filmstripCounter.is(":visible")) {
+                $(".bcc-detail-left").width($filmstripCounter.position().left-padding-buffer);
+            } else {
```
bceacuna: Is padding-buffer a variable? personally I prefer space around operators. for me it makes the code much more readable.

bceacuna: Actually, you are subtracting padding and buffer from $filmstripCounter.position().left.

bceacuna:
Looks good. Merge away

lib/bcgs_ford_modal.js
tyard: Yes. I will put spaces in -- a fair point. But you should also know that in JS, just like AS, you can't have a variable name with a hyphen (minus sign) in it, at least no using dot notation. So if you ever see object.property-value, you should never read property-value as a single variable name.

PR #15, opened 2015-06-18
Feature/disclaimer panel

bceacuna:
Implemented disclaimer panel without the custom scrollbar.


tyard:
Looks good. We will likely have to tweak the height of the overall modal because of this (the "openDrawerHeight", but otherwise looks fine to merge.

PR #16, opened 2015-06-19
Fix/save button

bceacuna:
The text SAVE is not part of the save button png.


lib/bcgs_ford_modal.js
```diff
@@ -154,6 +154,7 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
     //TODO: implement localization
     $detailDiv.find('.bcc-detail-right').html("Show Details &or;");
     $detailDiv.find('.bcc-detail-right').click(onShowHideDetails);
+    $detailDiv.find('.bcc-save-icon').html("Save");
```
tyard: I like this. It'll be easier to localize.

tyard:
Looks good.

PR #18, opened 2015-06-19
Code review fixes; separating carousel instantiation into its own file

BC-Mikey:
Code review changes for carousel. Separated carousel instantiation into its own file (bcgs_ford_filmstrip.js)


lib/bcgs_ford_filmstrip.js
```diff
@@ -0,0 +1,127 @@
...
+Copyright(c) 2015 Brightcove, Inc.
+
+Modal Window
+
+
+@author eacuna@brightcove.com
+@version 1.0.0
+
+v1.0.0: 6/2/2015
+TODO: remove instrinsic ratio implementation from index.html and it to the plugin
```
tyard: old comments?

lib/bcgs_ford_filmstrip.js
```diff
@@ -0,0 +1,127 @@
...
+            $carousel = $(".catalog-carousel");
+            $carouselContent = $("#bcc-carousel-content");
+            $modalContainer = $("#modal-container");
+            
+            if (!bcc.isMobile.any()) {
+                $nextBtn = new bcc.NavigationButton('next-btn', $carouselContent);
+                $prevBtn = new bcc.NavigationButton('prev-btn', $carouselContent);
+                $nextBtn = new bcc.NavigationButton('next-btn', $modalContainer);
+                $prevBtn = new bcc.NavigationButton('prev-btn', $modalContainer);
+            }
```
tyard: I am not following this. You are creating and then overwriting the $nextBtn and $prevBtn variables. I don't think you are even using these values, are you? If not, you don't need the assignments. Just do:

new bcc.NavigationButton('next-btn', $carouselContent);
new bcc.NavigationButton('prev-btn', $carouselContent);
new bcc.NavigationButton('next-btn', $modalContent);
new bcc.NavigationButton('prev-btn', $modalContent);

lib/bcgs_ford_main.js
```diff
@@ -74,8 +72,8 @@ v1.0.0: 6/2/2015
             bcc.unbind();
             playlist.unbind();
             
-            console.log("load carousel...");
-            loadCarousel(config.videos);
+            // TODO: does this belong here?
+            bcc.FilmStrip.loadCarousel(config.videos);
```
tyard: I think instead of coupling this with the main class and calling a specific method it would be cleaner to have the carousel plugin just listen for a change in the playlist.

With this, you can also get rid of the modalclose handler since you would just be listening and changing playlists whenever the playlist changed.

lib/bcgs_ford_filmstrip.js
```diff
@@ -0,0 +1,127 @@
...
+            
+            $(".next-btn").on('click', onNextBtn);
+            $(".prev-btn").on('click', onPrevBtn);
+            $modalContainer.on("modalClose", onModalClose);
+            
+            $content.on("catalogItemClick", function(event, data) {
```
tyard: As much as possible, let's define constants for events and avoid strings. There is already a bcc.events.SELECT -- maybe this could be used?

BC-Mikey: SELECT will work fine, I assume so long as it doesn't get used elsewhere in the project.

tyard: It is, but it shouldn't make a difference. You don't dispatch things globally, just from your object. So only those listening to your object would get notified.

lib/bcgs_ford_main.js
BC-Mikey: Makes sense. Will commit to this branch shortly.

lib/bcgs_ford_filmstrip.js
BC-Mikey: Thanks for the further clarification.

lib/bcgs_ford_filmstrip.js
BC-Mikey: Was going to try test that out, thanks for suggesting it!

lib/bcgs_ford_filmstrip.js
BC-Mikey: Yup, copied from Eric's boilerplate code. It's updated now.

BC-Mikey:
This branch has been updated. When ready it will be merged into branch #13 and then to dev.

PR #20, opened 2015-06-22
Refresh

tyard:


css/bcgs_ford_main.css
```diff
@@ -156,69 +160,136 @@
...
-@media only screen and (max-width: 568px) {
+@media only screen and (max-width: 767px) {
     #modal-container {
         margin: 0px;
-        width: 100%;
-        height: 100%;       
     }
+}
+
+@media only screen and (max-width: 640px) {
```
bceacuna: shouldn't we use max-device-width instead?

lib/bcc_core.js
```diff
@@ -219,6 +216,25 @@ v1.0.6: 6/17/2015
...
                     },
+                    
+                    /** 
+                     * Returns value from CSS as a number (in case it is a string with "px").
+                     */
+                    getCSSNumber = function(element, property) {
+                        // IE9+
+                        // @todo: IE8
+                        var value = window.getComputedStyle(element).getPropertyValue(property),
+                            num = parseInt(num, 10);
```
bceacuna: is this going to work on all supported browsers? getComputedStyle is only supported on ie11+

tyard: Where do you see that? The docs I have seen say IE9+, and you can see my note todo for IE8.

css/bcgs_ford_main.css
tyard: max-width is more responsive and allows us to easily test mobile layouts on the desktop by simply resizing. If we feel in the end that max-device-width gets us something better, we can switch once dev is complete, but I don't see harm in displaying the mobile layout if the user resizes to something that small. We do the same on the Met, I believe.

lib/bcgs_ford_modal.js
```diff
@@ -118,57 +125,97 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
+            $description.html(videoConfig.videoDescription || video.long_description);
             //shows/hides disclaimer panel based on player config
             if (disclaimer && disclaimer.trim().length > 0){
                 $disclaimerDiv.html(disclaimer);
-                $disclaimerDiv.show();        
+                $modal.addClass("bcc-has-disclaimer");
             } else {
-                $disclaimerDiv.hide();
+                $modal.removeClass("bcc-has-disclaimer");
             }
```
bceacuna: why $modal.removeClass("bcc-has-disclaimer") instead of $disclaimerDiv.hide()? is a hidden element clickable?

lib/bcgs_ford_modal.js
```diff
@@ -118,57 +125,97 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
+            findMarginAndPadding();
+            setModalHeight();
+            
             var aspectRatio = 16/9,
-                maxWidth = $(window).width()-margin*2,
-                maxHeight = $(window).height()-margin*2,
-                minHeight = maxWidth/aspectRatio+openDrawerHeight,
-                modalWidth = minHeight > maxHeight ? (maxHeight-openDrawerHeight)*aspectRatio : maxWidth,
+                modalWidth = Math.min(maxWidth, $(window).width()-margin*2),
                 videoHeight = modalWidth/aspectRatio,
```
bceacuna: Space around operators? For me it makes the code much more readable

bceacuna:
Looks good. Just a few comments/questions.

lib/bcc_core.js
bceacuna: http://caniuse.com/#search=getComputedStyle

tyard: I guess I am reading that chart completely differently. I see green boxes for IE9 and IE10, meaning they are supported. Do you see something different?

lib/bcgs_ford_modal.js
tyard: Controlling CSS through JS can be problematic. You can't then override with any other class -- the JS takes precedence. It's good to keep all of the styling in the CSS, even hiding elements.

lib/bcgs_ford_modal.js
tyard: OK

PR #22, opened 2015-06-22
reenabling carousel

tyard:
Mikey, I modified the buttons so that the ones over the player are vertically aligned and fade with the controls. I also made them go to the previous/next video as opposed to page.

I did notice that the images over the player are smaller than the close button. I think the images for the ones over the player should be a different size than the ones over the carousel, to match the PSD.


BC-Mikey:
Thanks Todd!

PR #24, opened 2015-06-23
Using larger carousel images; Resized buttons over video player

BC-Mikey:
I just missed getting this into #22 before it was merged to dev. This resolves the size of the carousel buttons that are over the video player. Once this is fixed I can update the staging page and send the following tickets over to Alfredo: FORD-18, FORD-31, among others.


css/bcgs_ford_carousel.css
```diff
@@ -115,11 +115,17 @@
 
 .bcc-next-btn {
     background: url('../img/carousel-next.png');
+    background-size: 100%;
     top: 65px;
+    /** IE5+ fix **/
+    filter: progid:DXImageTransform.Microsoft.AlphaImageLoader(src='../img/carousel-next.png', sizingMethod='scale');
```
tyard: Can you explain this?

tyard: nm -- just found reference. But using background-size:100% is equivalent to cover?

tyard:
Looks fine to merge (remember, just push to staging from a synced dev, though).

css/bcgs_ford_carousel.css
BC-Mikey: Yeah. It just says to scale it to be 100% to the height and width of the div. You can also say background-size: (width) (height);

i.e., background-size: 100% 50%;

Cover, I think, always makes the image as large as possible and could cut off parts of the image in some cases? I could be wrong about that.

BC-Mikey: I also only used the IE fix because 'background-size' is not supported in IE8, which is listed as a supported browser in the SOW.

tyard: Yeah, I think it would be helpful if the comment said IE8 instead of iE5, which we will never care about.

As for cover, it will scale the image up or down based on the size. If you don't keep the aspect ratio, yes, it could crop. That's generally what I've used for image scaling. If 100% works in this case, that's fine.

PR #27, opened 2015-06-23
Feature/ct adesktop

bceacuna:
Show CTAs on Desktop


lib/bcgs_ford_modal.js
```diff
@@ -129,16 +156,75 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
+            //Shows/Hides CTA container.
+            if (addedCTAs > 0) {
+                $ctaContainer.addClass("bcc-has-ctas");
+            } else {
+                $ctaContainer.removeClass("bcc-has-ctas");
+            }
+                 
+        },
+        
+        /**
```
tyard: To get this in we can leave it here for now. We have to be careful adding too much to a single file, though, before it gets really unwieldy. I think having the CTAs in their own file, like the filmstrip, would make this cleaner.

css/bcgs_ford_main.css
```diff
@@ -221,6 +221,29 @@
...
+    cursor: pointer;
+    position: relative;
+    display: inline-block;
+    margin-right: 2%;
+    width: 160px;
+    height: 45px;
+    background-color: #2d96cd;
+    color: white;
+    /*Aligns text*/
+    text-align: center;
```
tyard: I don't think we need this comment -- that text-align aligns text should be apparent, even if we didn't know what it was.

css/bcgs_ford_main.css
```diff
@@ -221,6 +221,29 @@
...
+}
+    .bcc-has-ctas.bcc-cta-container{
+        display: block;
+    }
+
+.bcc-cta-button  {
+    cursor: pointer;
+    position: relative;
+    display: inline-block;
+    margin-right: 2%;
```
tyard: Is there a reason for percent vs. pixel here? With resizing, the percents can be difficult at times, whereas 5px is always 5px. Just wondering in this case in particular if there was a specific need.

lib/bcgs_ford_modal.js
```diff
@@ -122,6 +124,31 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
         },
         
         /**
+        * Returns true if the argument text is a valid string, false otherwise
+        * TODO: move this utility function to core or web
```
tyard: Any reason these couldn't be put there immediately instead of adding a todo?

I think I would argue for a new bcc_util.js file if so. I haven't really had the need for these in any project, so it might be good to have a separate file with a number of helpful but non-necessary utils that we could bring in when needed.

lib/bcgs_ford_modal.js
```diff
@@ -129,16 +156,75 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
+                $ctaContainer = $detailDiv.find('.bcc-cta-container'),
+                totalCTAs = (ctaConfig) ? ctaConfig.length : 0;
+            
+            //Removes CTAs from the container before adding new ones,
+            $ctaContainer.empty();
+                       
+            for (i = 0; i < totalCTAs; i++) {
+             // Validates CTA configuration. 
+             // Exits loop if the number of added CTAs is greater to the max allowed.
+             if (addedCTAs >= CTA_MAX) break;
```
tyard: I think you could combine this with the bottom line of the loop instead of starting a whole new iteration, then breaking. Right at the bottom:

if (++addedCTAs >= CTA_MAX) break;

lib/bcgs_ford_modal.js
```diff
@@ -129,16 +156,75 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
+                     }
+                 }); 
+
+                 //Adds element to the display
+                 $ctaContainer.append($item);
+
+                 addedCTAs += 1;
+                 log(addedCTAs + ". Added CTA: " +  cta.text); //TODO: remove when testing is done
+
+             }());
```
tyard: This is fine. Just as a proposal of a optimized alternative: If you pass ctaConfig[i] to the function and then have cta as the argument, you can elimination several lines:

```
for (i = 0; i < totalCTAs; i++) {

    (function(cta) {
        var $item  = $('<div class="bcc-cta-button"></div>').html(cta.text);

        $item.on("click touchstart", function(){
            if (isValidFunction(cta.callback)){
                cta.callback(videoConfig); 
            } else {
                window.open(cta.link ,'_blank');
            }
        }); 

        //Adds element to the display
        $ctaContainer.append($item);
    }(ctaConfig[i]));

    if ($ctaContainer.children().length >= totalCTAs) break;
    
}
```

css/bcgs_ford_main.css
bceacuna: No, there was not a specific need. I'll change this to px.

css/bcgs_ford_main.css
bceacuna: fixed it

lib/bcgs_ford_modal.js
bceacuna: Yes, I wasn't sure core/web module was the right place for these utility methods. I'll include these methods to the bcc_util module next sprint

lib/bcgs_ford_modal.js
bceacuna: Done!

lib/bcgs_ford_modal.js
bceacuna: Good catch. I cleaned up this a little bit. Now I am passing ctaConfig[i] to the self executing anonymous  function.

lib/bcgs_ford_modal.js
bceacuna: Agree. Next sprint I'll refactor this and create a new file for the CTA logic.

PR #29, opened 2015-06-24
Feature/smartphone swiping

BC-Mikey:
Playlist swiping on a smartphone.


css/bcgs_ford_main.css
```diff
@@ -311,6 +311,7 @@
         top: 25px;
         left: 40px;
         font-size: 28px;
+        width: 100%;
```
tyard: Why? Is this not conflicting with the resize code (which is relying on the width being the width of the text)? I assume this is at least unrelated to the the swiping.

tyard: Ah, never mind -- see this will only apply to smartphone. All good.

lib/bcgs_ford_filmstrip.js
```diff
@@ -41,10 +41,17 @@ v1.0.0: 6/19/2015
...
+        playlist,
         $player = $("#player"),
         $carousel = $("<div class='bcc-carousel-content' />"),
         $modalContainer = $("#modal-container"),
         
+        // number of pixels a user must swipe in order to initiate a scroll on the playlist
+        X_PIXEL_DISTANCE_BEFORE_PAGESCROLL = 120,
+        
+        // number of milliseconds a page animation takes
+        SCROLLING_ANIMATION_TIME = 600,
```
tyard: unused?

lib/bcgs_ford_filmstrip.js
```diff
@@ -112,14 +119,94 @@ v1.0.0: 6/19/2015
          * load a carousel whenever the playlist model has been changed
          */
         onPlaylistChange = function(event) {
+            var videos;
+            
             if (event.data.property == bcc.model.PLAYLIST) {
-                var playlist = event.data.value,
-                    videos = playlist.getVideos();
+                console.log("CHANGE");
```
tyard: please remove

lib/bcgs_ford_filmstrip.js
```diff
@@ -112,14 +119,94 @@ v1.0.0: 6/19/2015
...
+                playlist = event.data.value;
+                playlist.bind(bcc.events.CHANGE, onPlaylistIndexChange);
+                videos = playlist.getVideos();
```
tyard: Not sure if I wrote this originally or you, but might as well just pass playlist.getVideos() to loadCarousel instead of the extra lines to declare and assign this variable.

lib/bcgs_ford_filmstrip.js
```diff
@@ -112,14 +119,94 @@ v1.0.0: 6/19/2015
...
+                carousel.index = playlist.getSelectedIndex();
+            }
         };
         
+        // TODO: duplicate code from carousel
+        $modalContainer.bind(
+            "touchstart",
+            function(e) {
+                var delta,
+                    pageChanging,
```
tyard: Just to prepare for generalizing this, let's use terms that apply. Here I think "swipeRegistered" could work.

lib/bcgs_ford_filmstrip.js
```diff
@@ -112,14 +119,94 @@ v1.0.0: 6/19/2015
...
+                    $modalPlayer = $("#modal-player"),
+                    // the smallest left/x position the video items should be placed at
+                    minX = Math.min(-($modalPlayer.width() - $modalContainer.width()), 0),
+                    // the starting positions on the swipe
+                    startX = parseInt($modalPlayer.css("left").substring(0, $modalPlayer.css("left").length-2), 10),
+                    startClickX = e.originalEvent.targetTouches ? e.originalEvent.targetTouches[0].pageX : e.clientX,
+                    
+                    /*
+                     * Handles updating the position of the video items during swipe
```
tyard: Comments can be modified as well.

lib/bcgs_ford_filmstrip.js
```diff
@@ -112,14 +119,94 @@ v1.0.0: 6/19/2015
...
+                     */
+                    touchmoveHandler = function(e) {
+                        var x = e.originalEvent.targetTouches ? e.originalEvent.targetTouches[0].pageX : e.clientX;
+                        delta = x - startClickX;
+                        $modalPlayer.css("left", Math.min(0, Math.max(startX + delta, minX)) + "px");
```
tyard: Are you actually moving the player?

lib/bcgs_ford_filmstrip.js
```diff
@@ -112,14 +119,94 @@ v1.0.0: 6/19/2015
...
         
+        // TODO: duplicate code from carousel
+        $modalContainer.bind(
+            "touchstart",
+            function(e) {
+                var delta,
+                    pageChanging,
+                    $modalPlayer = $("#modal-player"),
+                    // the smallest left/x position the video items should be placed at
+                    minX = Math.min(-($modalPlayer.width() - $modalContainer.width()), 0),
```
tyard: If you are not moving the player (later comment) then I don't think minX is needed.

lib/bcgs_ford_filmstrip.js
```diff
@@ -112,14 +119,94 @@ v1.0.0: 6/19/2015
...
+                        e.preventDefault();
+                    },
+
+                    /*
+                     * Handles updating the position of the video items after swipe, animating pages if approporiate
+                     */
+                    touchendHandler = function(e) {
+                        $modalContainer.unbind("touchmove touchend");
+                        // if there was enough of a swip, animate pages
+                        if (Math.abs(delta) > X_PIXEL_DISTANCE_BEFORE_PAGESCROLL) {
```
tyard: Perhaps constant should be "X_PIXEL_DISTANCE_BEFORE_SWIPE"?

lib/bcgs_ford_filmstrip.js
```diff
@@ -112,14 +119,94 @@ v1.0.0: 6/19/2015
...
+                            if (delta < 0) {
+                                // only proceed if we have a next page
+                                if (playlist.getSelectedIndex()+1 < playlist.getVideos().length) {
+                                    var selectedIndex = playlist.getSelectedIndex();
+                                    
+                                    pageChanging = true;
+                                    playlist.setSelectedIndex(selectedIndex+1);
+                                }
+                            }
+                            // means a swipe to the right, returning to previous page
```
tyard: Again, for refactor prep, I would suggest not having specific logic in here. Perhaps:

```
if (delta < 0) {
    swipeLeft();
} else {
    swipeRight();
}
```

Then, when we generalize, those lines are simply replaced with event triggers, and the code is already separated out in this file.

lib/bcgs_ford_filmstrip.js
```diff
@@ -112,14 +119,94 @@ v1.0.0: 6/19/2015
...
+                        };
+                        
+                        if (pageChanging) {
+                            e.preventDefault();
+                        }
+                    };
+
+                // we shouldn't do anything if there is only one page
+                // (note: should this be totalPages < 2? consider for future iteration)
+                if (pageChanging) return;
```
tyard: I think you can remove this.

PR #30, opened 2015-06-24
CTAs and layout on smart phones

tyard:


css/bcgs_ford_main.css
```diff
@@ -244,9 +241,10 @@
 .bcc-cta-button  {
     cursor: pointer;
     position: relative;
-    display: inline-block;
+    display: none;
     margin-right: 10px;
```
bceacuna: is it necessary to hide the bcc-cta-container's children as well?

tyard: This was to fulfill the functional requirements, using CSS. By default all the buttons are hidden. Only those with bcc-cta-desktop-show will appear on desktop. Only those with bcc-cta-mobile-show will appear on mobile. I was seeing things in the previous build where multiple buttons were appearing on smartphones.

lib/bcgs_ford_modal.js
```diff
@@ -160,55 +160,65 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
         /**
         * Adds CTAs to the displayer based on player configuration.
         */
         addCTAs = function(ctaConfig) {
             var i,
-                addedCTAs = 0, // number of CTAs added to the display
                 $ctaContainer = $detailDiv.find('.bcc-cta-container'),
-                totalCTAs = (ctaConfig) ? ctaConfig.length : 0;
+                totalCTAs = (ctaConfig) ? ctaConfig.length : 0,
+                smartPhone = isSmallScreen();
```
tyard: I don't think I am using this any more.

lib/bcgs_ford_modal.js
```diff
@@ -160,55 +160,65 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
...
+                    $ctaContainer.append($item);
 
-             }(ctaConfig[i]));
+                }(ctaConfig[i]));
                 
                 // Exits loop if the number of added CTAs is greater to the max allowed.
-                log(addedCTAs + ". Added CTA: " +  ctaConfig[i].text); //TODO: remove when testing is done
-                if (++addedCTAs >= CTA_MAX) break;               
+                if ($ctaContainer.children().filter(":visible").length >= CTA_MAX) break;               
             }
```
bceacuna: is it not more efficient having a counter instead of calling $ctaContainer.children().filter(":visible").length for each iteration? maybe it is micro optimization.

bceacuna:
Looks good. Just a few questions.

lib/bcgs_ford_modal.js
tyard: Counter can't be done here. What's happening is that all valid CTAs are being added. The CSS rules are determining which are visible and which are not. Only by checking the visibility each iteration can we know how many there are.

So if ALL the logic was in JS, then absolutely the counter would be better since the JS was controlling how many CTAs there are. But since the CSS is controlling the number of CTAs, we have to query using CSS. But this allows us to do things like toggle visibility of CTAs based on view, if we need to.

bceacuna: Why are we adding CTAs are not going to be displayed?

tyard: In the last iteration CTAs were being incorrectly added, not following the func spec and API, and I was trying to address that. In doing that I thought it was cleaner controlling what appeared and did not appear through the CSS. Again, what that allows is for logic that may or may not come up like "on landscape tablet, display all buttons, but in portrait only display the mobile-only buttons" and variations on that. Had to do that a lot for Met and thought I could prepare the files in a similar way here if it comes up. If it doesn't, I think controlling visibility through the CSS is still a clean way to go.

bceacuna: Thanks for the clarification.

bceacuna: I just noticed that you fully implemented https://bcconsulting.atlassian.net/browse/FORD-37. I was supposed to work on that task this week, maybe that's why you thought CTAa were not being correctly added.  I only implemented CTA's on desktop, and they were being added following the func spec.

Can we send FORD-37 to Alfredo?

tyard: Absolutely. Sorry about that. Alfredo logged a ticket to me for the display on smart phones with the CTAs a part of it (https://bcconsulting.atlassian.net/browse/FORD-57), so I thought they were supposed to be done since he was logging tickets. To resolve that I put in this code.

bceacuna: No problem, I'll assign the ticket to Alfredo.

PR #34, opened 2015-06-25
FORD-78: only swipe over video player; not over description area

BC-Mikey:


lib/bcgs_ford_filmstrip.js
```diff
@@ -161,15 +161,17 @@ v1.0.0: 6/19/2015
...
-            $modalContainer.bind(
+        if (bcc.isMobile.iPhone() || bcc.isMobile.AndroidPhone()) {
+            $player.bind(
                 "touchstart",
                 function(e) {
                     var delta,
                         swipeRegistered,
-                        $modalPlayer = $("#modal-player"),
+                        $player = $("#modal-player"),
                         // x position of where user starts to touch
```
bceacuna: $modalPlayer makes more sense to me. why did you rename this variable? Also note that there is another var called $player in the outer scope that is pointing to a different element in the DOM.

lib/bcgs_ford_filmstrip.js
```diff
@@ -161,15 +161,17 @@ v1.0.0: 6/19/2015
             }
```
bceacuna: Please remove TODO line 167
// TODO: needs Android Phone support as well

bceacuna:
Looks good, please merge after you make the necessary changes

PR #35, opened 2015-06-26
Fix/i phone scrolling

bceacuna:
iPhone does not allow you to scroll a modal window if its pointer-events property is set to auto. So to enable scrolling on iPhone, I had to  set pointer-events to none on the modal container. The only caveat with this fix is that now we need to set pointer-events to auto on each clickable element inside the modal container if we want then to listen for touch events.

I am setting pointer-events to auto on most of the clickable elements when the modal window is opened. for instance: 

    #modal-player.reveal .bcc-modal-close-button {
        pointer-events: auto;
    }


BC-Mikey:
This looks good to me! Merge away :)

PR #39, opened 2015-06-26
Short-term fix for FORD-81

BC-Mikey:


bceacuna:
looks good. Merge away.

PR #40, opened 2015-06-26
Feature/build file

bceacuna:
The build file adds the css/js minified files directly to css/lib folders when you run the default target. I also updated the index.html file, which is using the minified version of css/lib


BC-Mikey:
Hey Erick,

This is good. Does this mean that whenever we make changes from now on, we will need to run the grunt task and update them on our branches and update the respective compiled changes so they are reflected onto index.html?

bceacuna:
Yes, but you could also run "grunt watch" from your shell, so every time you update a css/js file, grunt automatically concatenates and minifies all the css/js files and moves the generated code to the build/ folder.

As you can see in this PR, the new index.html is loading the generated css/js files directly from the build folder. 

Don't forget to update the build file every time you add a *new* js/css file to the project.  JS libraries must be loaded in a specific order so make sure the new js are added at the end of the list

{code}
concat: {
      source: {
        src: ['lib/bcc_core.js', 'lib/bcc_web.js', 'lib/bcgs_ford_main.js', 'lib/bcgs_ford_modal.js', 'lib/bcc_carousel.js', 'lib/bcgs_ford_filmstrip.js', 'lib/bcgs_ford_carouselItems.js'],
        dest: names.concat
      },
{code}

BC-Mikey:
I know, I'm just talking more about having to update a branch with a minor fix in 1 file (e.g., bcgs_ford_filmstrip.js), but going forward we're going to have to make sure that the compiled files also reflect this.

For example, I just merged fix/ford-88 and I believe that's now missing from these compiled files. So should after this is merged, should I go back and add that fix to the grunt compiled files and re-merge?

bceacuna:
It is always a good idea test the final build. Note that the original index page is still in the project folder. I just renamed the file to index-dev.htlml. Feel free to use it.

BC-Mikey:
Ok cool. Merge away.

PR #41, opened 2015-06-29
FORD-77: Updating filmstrip counter when counter is not visible

BC-Mikey:
Fix for FORD-77, awaiting feedback from @tyard on the reasoning for only updating the $filmstripCounter while it's visible. This was the reason the number was not updating when switching a mobile device between portrait and landscape mode.


BC-Mikey:
Todd responded that the code removed in this commit needs to remain. Closing in favor of alternate solution.

PR #43, opened 2015-06-30
Highlight thumbnails when clicking next/prev buttons

BC-Mikey:
FORD-73


lib/bcgs_ford_filmstrip.js
```diff
@@ -134,10 +134,12 @@ v1.0.0: 6/19/2015
        
         onNextVideo = function(event) {
             bcc.trigger(bcc.events.REQUEST_NEXT_VIDEO);
+            carousel.highlightIndex();
         },
        
         onPrevVideo = function(event) {
             bcc.trigger(bcc.events.REQUEST_PREVIOUS_VIDEO);
+            carousel.highlightIndex();
         },
```
bceacuna: I was expecting to see highlightIndex() definitaion in this request. If highlightIndex() is already in place, merge away

BC-Mikey: What do you mean? I added it to both onPrevVideo() and onNextVideo()

bceacuna: My question is more around, Where is the highlightIndex method defined?

BC-Mikey: It's a method of the carousel class, so its part of bcc_carousel.js

bceacuna:
Looks good. Merge away.

PR #44, opened 2015-07-02
Feature/filmstrip toggle

BC-Mikey:
FORD-62: Implemented Filmstrip button into the controls as well as its ability to show/hide the Filmstrip.

TODOs are noted in the file including:
-correcting bcc.bind & bcc.trigger logic
-moving the filmstrip toggle handler into the FilmStrip class
-implementing the functionality of the Social Share button (FORD-64)


css/bcgs_ford_controls.css
```diff
@@ -0,0 +1,40 @@
+.bcc-player-button.bcc-filmstrip-control {
+    order: 5;
+    -ms-flex-order: 5;
+    -webkit-order: 5;
+    -webkit-box-ordinal-group: 5;
+    -moz-box-ordinal-group: 5;
+}
+
+    .bcc-player-button.bcc-filmstrip-control:before {
+        content: '';
```
bceacuna: When you use the before pseudo-class :before, the icon is not properly resized/position in some browsers/devices. Please verify if the button is properly rendered in iPhone  and Firefox.  Zoom in/out player by pressing CONTROL +/-. If everything works as expected you don't need to change this.

I ran into similar issues when I implemented a custom button for National Grid. This is what I did

https://github.com/BrightcovePS/National_Grid_CTA/blob/dev/html5/style/bcc_cta_plugin.css#L22

lib/bcc_web.js
```diff
@@ -78,12 +88,81 @@ v1.0.0: 10/29/2014
         loadVideo = function(vid) {
             cueVideo(vid);
             player.play();
-        };
+        },
+        
+        toggleFullscreen = function(node) {
```
bceacuna: I would recommend adding API docs to this method. The method name is self-explanatory but the argument needs to be briefly commented

lib/bcgs_ford_playercontrols.js
```diff
@@ -0,0 +1,144 @@
...
+        loadButtons = function() {
+            var filmstripBtn,
+                socialSharingBtn;
+            
+            if (filmstripLoaded) {
+                filmstripBtn = new bcc.Button(bcc.videoPlayer, FILMSTRIP_TOGGLE_CLASS, bcc.videoPlayer.controlBar.el(), filmstripToggleClick);
+            }
+            
+            socialSharingBtn = new bcc.Button(bcc.videoPlayer, SOCIAL_SHARING_CLASS, bcc.videoPlayer.controlBar.el(), socialShareClick);
+
```
bceacuna: I believe the sharing button should be enabled based on player configuration (socialNetworks). if socialNetworks.length = 0 the button shouldn't be added to the media controls. I can also add a method to the Social Module that you can use to enable/disable the button bcc.SocialPanel.enableButton(). We can think about the integration later. Just add a TODO in the meantime.

bceacuna:
Looks good. Merge away after making all the necessary changes.

thanks

lib/bcgs_ford_playercontrols.js
BC-Mikey: Ah! Thanks. Will add a TODO to this branch and take care of it in the new social-share button branch.

lib/bcc_web.js
BC-Mikey: Seems like all of the bcc_web methods are currently uncommented. I will make note of this for a change to the global PlayerLibrary.

css/bcgs_ford_controls.css
BC-Mikey: The buttons won't appear on an iPhone, however, they look to appear fine on Firefox when zooming in/out, aside from the expected edginess when enlarging PNG files of a small size...but this will be resolved once we start using the SVG sprite sheet. However, the placement of the '...' icon could be a little better so it is centered with the progress bar.

PR #48, opened 2015-07-06
Feature/social networks

bceacuna:
MIkey, 

I created the CSS SVG Sprite. The svgs are embedded in a css file, you can find this css file under img/csssvg. To display an svg, you just need to add one of the available .bcc-img-??????? (for instance .bcc-img-facebook) classes to an element. 

This implementation does not support Smart Phones yet.


css/bcgs_ford_social.css
```diff
@@ -0,0 +1,59 @@
...
+    background-color: #f7f7f7;
+}
+
+.bcc-social-icons {
+    /*removes gap between inline-blocks*/
+    font-size: 0px;
+}
+
+bcc-social-icons-row {
+}
```
BC-Mikey: Should this be deleted? This isn't a class and doesn't have any styles applied to it.

index.html
```diff
@@ -143,6 +150,71 @@
...
+                                    {
+                                        id: "pininteres",
+                                        icon: "http://icons.iconarchive.com/icons/danleech/simple/256/pinterest-icon.png",
+                                        callback: function(video) { console.log("pinterest: " + video.videoId); }
+                                    },
+                                    {
+                                        id: "pininteres",
+                                        icon: "http://icondesignfl.com/wp-content/themes/icon/images/icon_pin_large.png",
+                                        callback: function(video) { console.log("pinterest: " + video.videoId); }
+                                    }
```
BC-Mikey: Are you duplicating "pininteres" as a test of how many icons you can add? I think that the player only supports 4-11 icons, correct? (plus an e-mail icon)

Also, isn't it Pinterest? https://www.pinterest.com/

css/bcgs_ford_social.css
bceacuna: I believe this will be used to style the mobile version. I'll delete it if that is not the case.

lib/bcgs_ford_social.js
```diff
@@ -0,0 +1,276 @@
...
+            embedUrl = videoConfig.embedUrl;
+            
+            if(!bcc.isValidLink(shareUrl)){
+                shareUrl = window.location.href;
+            }
+            if (!bcc.isValidString(embedUrl)){
+                //TODO: Use player API go the embed code.
+                embedUrl = "<iframe>player code goes here.</iframe>";
+            }
+            console.log(player.contentEl());
```
BC-Mikey: remove console.log

index.html
bceacuna: Yes, this is just the testbet, I added all these icons only to test the reverse-wrapping functionality. Based on the requirements the player supports 0-12 icons (including e-mail). We are exposing an API, so Ford can add as many icons as they want up to 12. The original requirements changed a little bit. You can review the new requirements in JIRA.

https://bcconsulting.atlassian.net/browse/FORD-65

lib/bcgs_ford_social.js
bceacuna: Sure, I'll do it directly on this branch after you finish the review.

lib/bcgs_ford_social.js
```diff
@@ -0,0 +1,276 @@
...
+                window.open(url, '_blank');
+                //TODO: implement email
+            } else if (isBuiltin) {
+                log("Email icon was clicked");
+                if (bcc.isValidFunction(network.callback)) { 
+                    //TODO: dispatch event to open e-mail form
+                    log("open custom e-mail form");
+                } else {
+                    console.log("open native e-mail client");
+                    window.location.href = "mailto:?body=" + encodeURI(network.message);
```
bceacuna: I have already removed this console.log

BC-Mikey:
Very minor updates. Also wanted to note that we'll eventually move those SVGs into 1 big SVG Sprite Sheet with all icons inside of 1 file. (FORD-68) Other than that, looks good to merge.

bceacuna:
I added all the icons that you e-mailed me to the css-svg Sprite. Based on the article that I sent you last week, pure SVG Sprites are not well supported on Safari. 

FORD-68 has been partially implemented, we just need to add any missing icon to the css svg Sprite that we have already in place.

PR #49, opened 2015-07-06
Feature/player control styles

BC-Mikey:
@bceacuna this branch should resolve the filmstrip issue you saw on the iPad and also styles much of the player controls for FORD-63 (though it will still need some minor tweaks before I assign that ticket over to Alfredo).


css/bcgs_ford_controls.css
```diff
@@ -31,11 +62,20 @@
...
-        width: 27px;
-        height: 32px;
+        width: 30px !important;
+        height: 27px;
     }
     
     .bcc-player-button.bcc-social-control:hover:before {
         content: '';
         background: url(../img/social-on.png) no-repeat;
-    }
```
bceacuna: As I mentioned before. It was not trivial to add a new button to the media controls, the only way that I found to make it work on all the devices was removing the :before class and content property. Merge away  if this fix the outstanding issue.

PR #50, opened 2015-07-07
Committing fix

BC-Mikey:
Hey @bceacuna,

I created this branch because I realized that when using the 'nav buttons' to load in the next or previous video, this does not have the same effect as actually clicking on a thumbnail from the filmstrip. Previously, selecting the nav button only highlighted a thumbnail from the filmstrip. Now, however, I am triggering the SELECT event which takes care of the highlight and also updates the Carousel Object the same way that clicking on a thumbnail would.
-Mikey


bceacuna:
looks good. Merge away

PR #52, opened 2015-07-09
Feature/social mobile

bceacuna:
This branch contains the code for social panel (smart phones)


build/ford-custom-player.css
```diff
@@ -1,664 +0,0 @@
-.bcc-carousel {
-    background: #242424;
-    height: 200px;
```
bceacuna: In this branch, I removed all the generated code from the repo

build/ford-custom-player.js
```diff
@@ -659,7 +659,17 @@ v1.0.0: 7/01/2015
```
bceacuna: You don't need to review this file. I am not sure why it is not all red. I added build/ to .gitignore

build/ford-custom-player.css
BC-Mikey: Makes sense!

css/bcgs_ford_main.css
```diff
@@ -12,6 +12,13 @@
          url('../fonts/regular/extralight.svg#webfont') format('svg');
 }
 
+/**
+* Prevents scrolling the main content when the modal window is opened (workaround for ios)
+
+body.noscroll{
+    overflow:hidden;
+}*/
+
```
BC-Mikey: Why is this commented out? Should it just be removed?

css/bcgs_ford_main.css
```diff
@@ -200,30 +208,32 @@
...
-.bcc-save-icon {
-    cursor:pointer;
-    display: none;
+.bcc-detail-buttom-container{
     position:absolute;
-    height: 42px;
+    right: 0px;
     top: 130px;
-    right: 60px;
+    right: 60px;
```
BC-Mikey: You have right defined twice here. Also, you called the class 'buttom-container'. Should this be 'bottom' or 'button'?

css/bcgs_ford_main.css
bceacuna: I'll remove it

css/bcgs_ford_main.css
```diff
@@ -340,28 +359,44 @@
...
+        font-size: 14px;
+        line-height: 21px;
         padding-top: 0px;
     }
+    
+    .bcc-detail-buttom-container{
+        top: 20px;
+        right: 20px;  
+        line-height: 10px;
+    }
```
BC-Mikey: Using 'buttom' here again. Please change to either 'button' or 'bottom'

css/bcgs_ford_main.css
bceacuna: Good catch. I'll fix this.

css/bcgs_ford_main.css
bceacuna: I'll fix this type

lib/bcgs_ford_social.js
```diff
@@ -245,7 +257,8 @@ TODO: Add embed code to the social panel.
             
     //Adds listener to the social icon container
     $iconsDiv.on("click touchstart", onIconContainerClick);  
-    $closeButtonDiv.on("click touchstart", onCloseButtonClick);        
+    $closeButtonDiv.on("click touchstart", onCloseButtonClick); 
+    $xCloseDiv.on("click touchstart", onCloseButtonClick);
```
BC-Mikey: $xCloseDiv and $closeButtonDiv appear to do the same thing here, which I'm guessing is closing out the social panel, is that correct? Or is $closeButton another 'X' icon that lives behind the social panel? This is confusing me a bit and I think it has to do with the naming and variables you've chosen...

It seems that $closeButtonDiv exists as an X button when the social menu is not open:
$closeButtonDiv = $('<div class="bcc-modal-close-button" />'),

However, is supposed to represent a different div (a textual 'Close' representation) when it is open:
$closeButtonDiv = $socialDiv.find('.bcc-close-button'),
$closeButtonDiv.html("Close");

It would make more sense for the $closeButtonDiv to always remain the 'X' but then the social menu having its own specific variable for the textual representation. Does that make sense?

BC-Mikey:
Just a few comments. Once looked over can be merged.

lib/bcgs_ford_social.js
bceacuna: $xCloseDiv is the X icon that is on the upper right corner of the social panel. $closeButtonDiv is the button at the bottom of the social panel, which is labeled 'Close'. Both buttons do the same, they close the social panel. The 'X' button is only shown in smart phones. I believe that is the confusion.  

Does that make sense?

BC-Mikey: Oh, got it. I was thinking this $closeButtonDiv was full X button that closes the entire modal when viewing the player on the Desktop (without the social menu open).

PR #57, opened 2015-07-13
Removed margins of custom buttons in favor of custom spacers to resol…

BC-Mikey:
…ve inactive progress bar style issue

Progress bar now displays in full inactive state with this fix. However, fake "spacer" buttons are added instead of using margins to style the custom buttons.


lib/bcgs_ford_playercontrols.js
```diff
@@ -49,22 +49,28 @@ v1.0.0: 7/2/2015
...
          */
         loadButtons = function() {
             var filmstripBtn,
                 socialSharingBtn;
             
             if (filmstripLoaded) {
+                new bcc.Button(bcc.videoPlayer, SPACER_CLASS, bcc.videoPlayer.controlBar.el());
                 filmstripBtn = new bcc.Button(bcc.videoPlayer, FILMSTRIP_TOGGLE_CLASS, bcc.videoPlayer.controlBar.el(), filmstripToggleClick);
+                new bcc.Button(bcc.videoPlayer, SPACER_CLASS, bcc.videoPlayer.controlBar.el());
             }
```
bceacuna: How do you prevent not to highlight the spacer when you tab? I believe every new element that is added to the media controls are added to the tab order .

BC-Mikey: That's a good point... I haven't looked much into the Accessibility yet or know much about it. I know it's a task for later and I will have to familiarize myself with it. Is Accessibility controlled with aria attributes and roles? If so, we could remove those from the spacer, or create a subclass separate from bcc.Button() specific to creating the spacers.

bceacuna: The tab order is set using tabindex. I believe you can exclude an element from the tab order by setting its tabindex="-1". Try it I am not sure if it works across all browsers.

BC-Mikey: Tough to tell if it's working... they don't exactly have highlight states. Though in Safari/Chrome I can tab -> tab+enter will play/pause, tab will highlight the progress bar, tab+enter again will enter/exit full screen. Firefox there seems to be a lot more tabs in between these buttons, including the Volume which didn't work in the other browsers? Perhaps this will need more investigation on when we dive into the Accessibility, but for now I can just remove the tabindex from these div's altogether?

BC-Mikey:
Hey @bceacuna,

I've updated this to remove the tabindex and other accessibility related attributes from the spacer elements. Things look like they're working fine. In some browsers the tabbing order of the volume and full screen is switched, but that just looked like it's because of how they're placed in the DOM. We have a larger story to work on the player's accessibility so we can tackle that then.

I also made a TODO for making a Spacer component, either a class, or something to add to bcc-util.js eventually. For now, as a quick fix I'm keeping these as buttons and stripping the accessibility attributes.

-Mikey

bceacuna:
Ok. WE can revisit this  later. Merge away.

PR #59, opened 2015-07-13
FORD-77

tyard:
ensure when orientation changes that the filmstrip number is properly represented


BC-Mikey:
Merged Todd's changes for FORD-77

PR #60, opened 2015-07-13
FORD-76, FORD-91

tyard:
When hiding player, don't set display to none, which throws exception in player.


BC-Mikey:
Looks good; merging

PR #62, opened 2015-07-15
Feature/email form

bceacuna:
This PR contains the custom e-mail form for desktop only. You can see it working on my dev environment


.gitignore
```diff
@@ -51,3 +51,13 @@ node_modules/
...
+
+build/ford-custom-player.min.js
+
+build/ford-custom-player.js
+
+build/ford-custom-player.js
+
+build/ford-custom-player.min.js
+
+build/ford-custom-player.min.css
```
bceacuna: for some strange reason .gitignore was not working as expected. The build/ folder was added to .gitignore long time ago but the generated code was not removed from the repo

css/bcgs_ford_email.css
```diff
@@ -0,0 +1,114 @@
...
+    margin-bottom: 15px;
+    padding-left: 15px;
+    padding-right: 15px
+}
+
+
+.bcc-email-form-from {
+    display: inline-block;
+    padding-right: 25px;  
+}
```
BC-Mikey: Right now, when the 'Invalid e-mail' text is added to the form if the user hits send, the form pushes down because the SPAN tag is set to display:inline.

I recommend setting both '.bcc-email-form-from' and '.bcc-email-form-to' to include 'position:relative'. This way, you can set 'position: absolute' on your SPAN tag and place as needed relative to their display boxes, without causing them to shift down. (e.g., top: -20px)

lib/bcgs_ford_email.js
```diff
@@ -0,0 +1,226 @@
...
+                isToValid = validateEmail(to); 
+            if (isFromValid){
+                $fromError.hide();               
+            } else {
+                $fromError.show();                
+            }
+            if (isToValid){              
+                $toError.hide();
+            } else {
+                $toError.show();
```
BC-Mikey: This is currently pushing the form down. Please see my above comment about adding the following styles to it:

.bcc-email-form-from {
    position: relative;
}
.bcc-email-form-to {
    position: relative;
}
.bcc-email-form-from > span, .bcc-email-form-to > span {
    position: absolute;
    top: -20px; /* or whatever spacing you need */
}

css/bcgs_ford_email.css
bceacuna: Good point. I'll add this to me todo list.

PR #64, opened 2015-07-15
Feature/email form rev

bceacuna:
Mikey, 

The form is not pushed down anymore. The error message is added under the textfield. It looks much better. 


BC-Mikey:
Hey Erick, I see that you added position: relative to the container divs, but where is the position:absolute for the 'span' DIV?

css/bcgs_ford_email.css
```diff
@@ -18,14 +18,16 @@
...
 .bcc-email-form-textfields{ 
-    margin-top: 40px;
-    margin-bottom: 10px;
+    margin-top: 35px;
+    margin-bottom: 15px;
     width: 100%;    
 }
     .bcc-email-form-textfields span {
+        position: absolute;
         display: none;
```
bceacuna: My latest change was not saved. This is the line that you were looking.

BC-Mikey:
Thanks Erick! Looks good!

PR #69, opened 2015-07-17
fixing merge so exceptions no longer thrown

tyard:


lib/bcgs_ford_disclaimer.js
```diff
@@ -52,6 +52,12 @@ v1.0.0: 6/19/2015
         setScrollingValues = function() {
             textHeight = $disclaimerText.height() + bcc.getCSSNumber($disclaimerText[0], "padding-top")*2;
             areaHeight = $(".bcc-detail-disclaimer-panel").height();
+            console.log(textHeight);
```
tyard: will remove

lib/bcgs_ford_filmstrip.js
```diff
@@ -235,11 +235,11 @@ v1.0.0: 6/19/2015
         
         bcc.bind(bcc.events.INITIALIZED, function() {
             bcc.model.bind(bcc.events.CHANGE, onPlaylistChange);
+            bcc.bind(bcc.events.FILMSTRIP_TOGGLE_CLICK, onFilmstripToggle);
```
tyard: moved this here as class should handle its own binding, not be within main

lib/bcgs_ford_social.js
```diff
@@ -243,6 +243,7 @@ TODO: (Smart Phone) Implement animation to reveal the sharing selection options.
     //set event listeners
     bcc.bind(bcc.events.INITIALIZED, function() {
         bcc.model.bind(bcc.events.CHANGE, onModelChange);
+        bcc.bind(bcc.events.SOCIAL_SHARE_CLICK, openPanel);
```
tyard: moved this here as class should handle its own binding, not be within main

lib/bcgs_ford_main.js
```diff
@@ -77,8 +78,14 @@ v1.0.0: 6/2/2015
...
             // clear previous handlers
-            bcc.unbind();
             playlist.unbind();
+            // @note: Please do NOT clear all handlers from bcc using bcc.unbind().
+            // This is done with the playlsit because the playlist is being destroyed.
+            // bcc is still active and shouldn't have all of its listeners removed, 
+            // as was done in a previous commit.
+            bcc.unbind(bcc.events.REQUEST_NEXT_VIDEO);
+            bcc.unbind(bcc.events.REQUEST_PREVIOUS_VIDEO);
+            bcc.unbind(bcc.events.REQUEST_VIDEO_INDEX);
```
tyard: Please be REALLY careful with this -- took me a while to see why my listeners for bcc were disappearing. unbind() on playlist was being done because the playlist model was being destroyed and recreated (which occurs for each new open of the panel). bcc, however, should remain constant -- there's no reason to re-add all handlers to that every time the panel is opened.

lib/bcgs_ford_main.js
```diff
@@ -105,10 +112,6 @@ v1.0.0: 6/2/2015
                     playlist.setSelectedIndex(index);
                 }
             });
-            bcc.bind(bcc.events.FILMSTRIP_TOGGLE_CLICK, bcc.FilmStrip.onFilmstripToggle);
-            bcc.bind(bcc.events.SOCIAL_SHARE_CLICK, function(event) {
-                bcc.SocialPanel.open();
-            });
             // add playlist to main model
```
tyard: Moved these to separate classes as I couldn't see a good reason to have those concerns here, nor the need to reestablish listeners every time. Isn't once enough?

lib/bcgs_ford_filmstrip.js
```diff
@@ -235,11 +235,11 @@ v1.0.0: 6/19/2015
...
         });
         $carousel.on(bcc.events.SELECT, requestVideo);
             
         return {
-            onFilmstripToggle: onFilmstripToggle
```
BC-Mikey: Thanks. Don't remember why I added this the way I did...

BC-Mikey:
The fixes make sense. Will be careful not to fuss with the binds & unbinds.

PR #71, opened 2015-07-17
Feature/email mobile

bceacuna:
This PR contains all the styles for the email sharing dialogue (smart phones), and also the logic to support a CSV of recipients. This resolves the problem involving the email panel being cut off and scrolling on portrait.  However, at the moment, scrolling on landscape does not work. I can look into this later today/next week, but I wanted to get this in for code review.


css/bcgs_ford_email.css
```diff
@@ -84,6 +85,13 @@
     font-size: 16px;  
     padding-top: 20px
 }
+    .bcc-email-form-video-title::before{
+        content: '"';
+    }
+    
+    .bcc-email-form-video-title::after{
+        content: '"';
+    }
```
BC-Mikey: Are you replacing the content with a double-quote here or do you mean to erase it, i.e.:

content: '';

css/bcgs_ford_email.css
```diff
@@ -116,3 +124,35 @@
...
+        width: 100%;   
+    }
+    
+    .bcc-email-form-details {
+        width: 100%;
+    }
+    
+    .bcc-email-container textarea {
+        width: 100%;
+    }
```
BC-Mikey: since these above 5 all have the same width: 100% style, can they all be condensed to:

.bcc-email-container input, .bcc-email-form-from, .bcc-email-form-to, .bcc-email-form-details, .bcc-email-container textarea {
width: 100%;
}

You might even be able to get away with adding .bcc-email-form-send and .bcc-email-form-cancel, and adding box-sizing and border-box to all of these divs.

lib/bcgs_ford_email.js
```diff
@@ -182,31 +204,72 @@ v1.0.0: 7/01/2015
...
 
         /**
         * Opens email panel.
         */
-        openPanel = function(){
-            //TODO: add $emailDiv.get(0) to $modal for smart phones
-            player.el().appendChild($emailDiv.get(0)); 
+        openPanel = function(){ 
+            isPanelOpen = true;
+            addPanel()
```
BC-Mikey: Not an error, but missing semicolon here. Just add for consistency.

css/bcgs_ford_email.css
bceacuna: This adds quotes to the video title, so if the title is Ford Motors after applying these pseudo classes,  "Ford Motors" is displayed.

BC-Mikey: Oh, got it. This isn't something we want to add through JS?

bceacuna: We can decorate the video title either way. You can use :before/:after to insert something before or after the content of an element.  I don't have any preference. I just thought it was a good way to add quotes to the video tittle

css/bcgs_ford_email.css
bceacuna: Done!

lib/bcgs_ford_email.js
bceacuna: Done!

css/bcgs_ford_email.css
BC-Mikey: Sure. I was wondering if this is how it should be done going forward and if we should adopt this as a dev process we use consistently for this type of behavior.

bceacuna: good question, I am not sure this is a best practice, but I have seen using :before and :after to decorate the content as I did it here.

tyard: Let's not "set it" just yet. Use either method. If we find there is a good reason for one over the other we will start setting it as standard.

PR #72, opened 2015-07-18
hide fullscreen button ie8/9/10

bceacuna:
This is the fix for FORD-110. Since the Brightcove in-page embed player does not support fullscreen on ie8/9/10, we decided to hide the fullscreen button when ie8-ie10 is detected.


lib/bcc_util.js
```diff
@@ -53,6 +53,23 @@ v1.0.0: 7/01/2015
...
+          var rv = -1; // Return value assumes failure.
+          if (navigator.appName == 'Microsoft Internet Explorer')
+          {
+            var ua = navigator.userAgent;
+            var re  = new RegExp("MSIE ([0-9]{1,}[\.0-9]{0,})");
+            if (re.exec(ua) != null)
+              rv = parseFloat( RegExp.$1 );
+          }
+          return rv;
+        },
```
tyard: Make sure that when updating common files you update the version and comments at the top.

bceacuna: Done!

PR #75, opened 2015-07-20
adding config to model, autoplay

tyard:


bceacuna:
Looks good.

tyard:
It's adding it to the main model, yes. We could add each value individually, but this is the simplest way of passing the values around.

PR #76, opened 2015-07-20
Added config parameter to toggle filmstrip expand/collapse

BC-Mikey:
Added 'defaultFilmstripExpanded' configuration parameter


lib/bcgs_ford_main.js
```diff
@@ -91,7 +91,8 @@ v1.0.0: 6/2/2015
             // create new playlist model from config video
             playlist = new bcc.Model({
                 selectedIndex: selectedIndex,
-                videos: config.videos
+                videos: config.videos,
+                defaultFilmstripExpanded: config.defaultFilmstripExpanded
```
tyard: My other commit added the config to the main model itself. I'd like to go with that as there are a lot of parameters we'll have to pass around and that keeps them in one place. Also, this playlist model is intended more to be an abstract collection of videos, not tied with UI. I think it will be cleaner.

Really, the only change would be to remove this line here, then to have bcc.model.getConfig().defaultFilmstripExpanded above.

BC-Mikey: Ah okay. Thanks!

tyard:
Looks good.

PR #77, opened 2015-07-20
added config for displaying filmstrip toggle button

BC-Mikey:
Here is the 2nd config param for displaying the filmstrip toggle button


css/bcgs_ford_controls.css
```diff
@@ -39,6 +39,10 @@
     content: '' !important;
 }
 
+.vjs-progress-control {
+    padding-right: 16px !important;
```
tyard: Any way to achieve this without !important? I have found this has caused issues in the past with responsiveness as it  can never be overridden. Is there a reason it is required in this case?

css/bcgs_ford_controls.css
```diff
@@ -109,16 +113,10 @@
     .vjs-spacer, .bcc-filmstrip-control, .bcc-social-control {
         display: none !important;
```
tyard: Hadn't noticed all these other instances as well. In general I think we want to avoid !important as much as possible in the future.

lib/bcgs_ford_playercontrols.js
```diff
@@ -64,7 +64,6 @@ v1.0.0: 7/2/2015
                 socialSharingBtn;
             
             if (filmstripLoaded) {
-                new bcc.Button(bcc.videoPlayer, SPACER_CLASS, bcc.videoPlayer.controlBar.el());
                 filmstripBtn = new bcc.Button(bcc.videoPlayer, FILMSTRIP_TOGGLE_CLASS, bcc.videoPlayer.controlBar.el(), filmstripToggleClick);
```
tyard: What's the point of this filmstripBtn var? I don't see it actually used anywhere.

tyard:
I'm not sure I understand this commit. Isn't the config supposed to be "showFilmstripToggle" and can't this change from each launch of the modal?

As such, I think it might be better to have a look at the model.getConfig() and simply add or remove a class from the modal like "bcc-has-filmstrip-toggle" and then set up the CSS to show or hide the control based on that. Same thing for the share button (which looks like is also outstanding).

Or am I missing the point of this commit?

css/bcgs_ford_controls.css
BC-Mikey: It does work without !important. It's just that the padding is not immediately there. So when the player first loads, the control bar is pushed right up against the closest button for a few seconds and then it fixes itself.

css/bcgs_ford_controls.css
BC-Mikey: Yeah, I've only added it in cases where I had to override styles that were inline or were loaded somewhere later by the default code. In the case of the buttons, the .vjs-button styles couldn't be overridden without !important

lib/bcgs_ford_playercontrols.js
BC-Mikey: Will remove.

css/bcgs_ford_controls.css
tyard: Did you try the tricks to make your CSS super-specific (like video-js.video-js)?

BC-Mikey: I just tried this and while it did work, the same issue appears to be happening where the styles aren't taking effect immediately upon load. In case of the custom buttons, this actually means that it takes a second or two of playback before the button graphics appear.

BC-Mikey:
I've added the showFilmstripToggle back. Not sure how it got removed. I think because I started making changes before merging the other config param PR.

PR #78, opened 2015-07-21
Fix/api

bceacuna:
This PR contains the fix for the player API. Now the implementation matches the API docs.


BC-Mikey:
Looks good Erick

lib/bcgs_ford_playercontrols.js
```diff
@@ -115,7 +115,7 @@ v1.0.0: 7/2/2015
             var playlist,
                 videos;
             
-            if (bcc.settings.showShareButton === false) {
+            if (bcc.model.getConfig().showShareButton === false) {
```
tyard: Is this going to work? This sets displayShareButton when the model changes, but the button is only added (or not added, depending on the value) once, because after that loadButtons() is not called a second time. E.g., if this is false the first time, then the button is not added. If it's true the next launch, this value is not actually ever set to true AND loadButtons() is not called a second time.

I think a better way of approaching this is to have only one (not two) classes look to this config value. Have a CSS class added to modal based on the value. Always add the button (so you don't need that conditional to add it or not), but use CSS to show or hide based on the class value.

lib/bcgs_ford_social.js
```diff
@@ -139,6 +146,33 @@ TODO: (Smart Phone) Implement animation to reveal the sharing selection options.
...
+        * Renders panel's configurable elements based on player configuration
+        */
+        configPanelView = function(){
+            //TODO: Verify if the social panel is recreated when player is relaunched 
+            // with a new configuration object. Test this after ford-76 is fixed.
+            
+            //Adds social icons to the panel       
+            addIcons(modalConfig.sharingOptions);
+            // Removes embedCode based on player configuration        
+            if (modalConfig.showEmbed === false){
```
tyard: I'd love to see these controlled by CSS instead of all these lines to manually show/hide.

lib/bcgs_ford_playercontrols.js
bceacuna: No, this is not going to work. I didn't spend any time on this module. I was plaining to ask Mikey to fix this after merging since he has been implementing the media controls

lib/bcgs_ford_social.js
bceacuna: I'll add a TODO. I'll cleaned up this next sprint or this week if I have time. Today and tomorrow I want to focus on the new tasks that were assigned to me.

PR #80, opened 2015-07-21
Fix/toggle styles

BC-Mikey:
Using styles to control the toggling of the configurable player buttons display.
I also added back the styles from Ford-76 into this PR since this is a small change.


css/bcgs_ford_controls.css
```diff
@@ -55,6 +57,8 @@
     -moz-box-ordinal-group: 5;
 }
 
+    .vjs-fullscreen .bcc-filmstrip-control { display: none; }
+
```
bceacuna: new line after curly bracket?

BC-Mikey:
Resolved line breaks after curly brackets

lib/bcgs_ford_playercontrols.js
```diff
@@ -111,20 +104,28 @@ v1.0.0: 7/2/2015
...
+                // toggle buttons based on state
+                if (videos.length < 1 || !bcc.model.getConfig().showFilmstripToggle) {
+                    hideButton(FILMSTRIP_TOGGLE_CLASS);
+                }
+                
+                if (!bcc.model.getConfig().showShareButton) {
+                    hideButton(SOCIAL_SHARING_CLASS);
+                }
             }
+        },
```
bceacuna: is this going to work? Try this.
1. launch the modal window with showShareButton set to false, share button is hidden,
2. close the modal window
3. re-open  the modal window again but this time with showShareButton set to true.
4.  share button is not shown in the media controls

lib/bcgs_ford_playercontrols.js
```diff
@@ -111,20 +104,28 @@ v1.0.0: 7/2/2015
...
+        
+        hideButton = function(buttonClass) {
+            $('.' + buttonClass).toggleClass('vjs-hidden');
+            $('.' + buttonClass).next().toggleClass('vjs-hidden');
         };
```
bceacuna: I am not very familiar with the vjs-hidden class. Can we use vjs-hidden to hide any element within the player or it is only used for elements in the media controls?

lib/bcgs_ford_playercontrols.js
tyard: Yes, I see the same. Can the buttons ever be revealed?

lib/bcgs_ford_playercontrols.js
BC-Mikey: It's a simple style from the default Video JS CSS that can be added anywhere. It's only property is display: none;

lib/bcgs_ford_playercontrols.js
BC-Mikey: Ah you're right. toggleClass() wouldn't ever be hit in this. I'll change this so that instead of using toggleClass we use addClass/removeClass, respectively.

BC-Mikey: Question: How are the config option values updated without reloading the page? (i.e., when would this be an issue and how do I test)

Right now I'm attempting to load the page with showShareButton set to false, I close out of the modal, use the Inspector to load up the DOM and edit the config options so that showShareButton is set to true. The 'onModelChange' event fires again, however, my config option always keeps the value that was initially set. It doesn't look like the configs are being updated.

tyard: You can add a second button to the index page that calls launchModal with a different set of parameters. Maybe we should have two or three buttons with different configs. Ideally, we would have a page with a form for setting the desired options (what can be set -- callbacks would be tricky), but simplest case would be three buttons with hardcoded params.

BC-Mikey: Got it thanks

BC-Mikey:
Sorry, I didn't create a rev/ branch for the comments re:
1. launch the modal window with showShareButton set to false, share button is hidden,
2. close the modal window
3. re-open the modal window again but this time with showShareButton set to true.
4. share button is not shown in the media controls

My mistake. I'll be sure to do that in the future. In the meantime, is there anything else that needs to be revised?

tyard:
looks good

lib/bcgs_ford_playercontrols.js
```diff
@@ -111,20 +104,37 @@ v1.0.0: 7/2/2015
...
+                
+                // toggle buttons based on state
+                if (videos.length < 1 || !bcc.model.getConfig().showFilmstripToggle) {
+                    hideButton(FILMSTRIP_TOGGLE_CLASS);
+                }else {
+                    showButton(FILMSTRIP_TOGGLE_CLASS);
+                }
+                
+                if (!bcc.model.getConfig().showShareButton) {
+                    hideButton(SOCIAL_SHARING_CLASS);
```
bceacuna: Here share button is not added to the media controls if showShareButton is not set, and it should .showShareButton is an optional configuration parameter that defaults to true if it is not provided. Try something like this to show the share button if showShareButton = true or it is not set.

if (bcc.model.getConfig().showShareButton === false){
  hideButton(SOCIAL_SHARING_CLASS);
} else {
  showButton(SOCIAL_SHARING_CLASS);
}

BC-Mikey: Thanks Erick!

How about the Filmstrip conditional. Basically, the filmstrip toggle needs to be hidden by default if there is only one video. Otherwise, the true/false config can be used to display the toggle button. If a playlist is present then it defaults to true.

Looking at it again, I should change it to:
if (videos.length <= 1 || bcc.model.getConfig().showFilmstripToggle === false) {

bceacuna: You have it right.

PR #82, opened 2015-07-21
Feature/override social

bceacuna:
This PR contains the changes to allow default social network values to be overridden. I also included a quick fix for the e-mail form. maxSharees was not being properly set.


BC-Mikey:
The updates make sense and look good to me.

tyard:
agreed, looks good

PR #83, opened 2015-07-22
Added new conditions for button toggle states

BC-Mikey:
In a closed PR, Erick pointed out that my conditionals didn't cover uses cases for when the param was not included at all. Also, I noticed my filmstrip toggle conditional may not have fully covered the following criteria:

"If there are no extra videos, this parameter should be ignored and the button should not be displayed."


bceacuna:
Looks good.

PR #85, opened 2015-07-22
show poster on start and end

tyard:


lib/bcgs_ford_playercontrols.js
```diff
@@ -137,19 +150,27 @@ v1.0.0: 7/2/2015
...
+        
+    //fallback workaround to fullscreen not working in IE9/10 
+    // (hides fullscreen button)
+    if (ieVersion < 11.0 && ieVersion > -1){    
+        $('.vjs-fullscreen-control').css("display","none");
+        //move volume panel to the right.
+        $('.vjs-menu .vjs-menu-content').css({"left": "30px", "z-index": "800"});
+    }
     
+    // return all public methods here
```
tyard: all this and below had been removed -- not sure why. just putting it back in to match our other files

lib/bcgs_ford_playercontrols.js
```diff
@@ -137,19 +150,27 @@ v1.0.0: 7/2/2015
...
-        });
-            
-        //fallback workaround to fullscreen not working in IE9/10 
-        // (hides fullscreen button)
-        if (ieVersion < 11.0 && ieVersion > -1){    
-            $('.vjs-fullscreen-control').css("display","none");
-            //move volume panel to the right.
-            $('.vjs-menu .vjs-menu-content').css({"left": "30px", "z-index": "800"});
-        }            
+    bcc.bind(bcc.events.INITIALIZED, function() {
```
tyard: just altering the indent to match format of other files

lib/bcgs_ford_playercontrols.js
```diff
@@ -98,32 +94,49 @@ v1.0.0: 7/2/2015
         },
         
         /**
+         * Switches to custom end poster, if required
+         */
+        changeToEndPoster = function() {
+            var video = bcc.model.getVideo();
+            log("here");
```
tyard: whoops, will remove

css/bcgs_ford_main.css
```diff
@@ -104,6 +104,9 @@
 #modal-player.reveal .video-js {
     pointer-events: auto;
 }
+    #modal-player .video-js.vjs-ended .vjs-poster {
+        display: block;
```
bceacuna: Where can I find the documentation for all there video-js classes?

bceacuna:
Looks good.

lib/bcgs_ford_playercontrols.js
BC-Mikey: Ah I removed it when I created the file because I wasn't returning anything... I'll be sure to leave this code in as part of the plugin template from now on.

css/bcgs_ford_main.css
BC-Mikey: I wish I knew, too. I find that most of these classes added to the main .video-js DIV describe the current state of the player. It may be useful to always first check the div if there is an existing class you can use to style a specific state. (.vjs-user-inactive, .vjs-user-active .vjs-fullscreen, .vjs-has-started, etc., etc.)

tyard: I am not sure they are listed anywhere. It is just something I found today as I was looking for the poster to reappear at the completion of the video. Maybe we can start our own list and add to it over time. Good thing for the backlog.

PR #87, opened 2015-07-23
close button opacity set to 1 if native player

bceacuna:
This PR contains the fix for FORD-103 ([Mobile] Close button is missing after full screen video player is closed)


tyard:
looks good

PR #90, opened 2015-07-24
FORD-115: highlight and display currently selected video on player load

BC-Mikey:


lib/bcgs_ford_filmstrip.js
```diff
@@ -87,6 +87,13 @@ v1.0.0: 6/19/2015
                 if (!bcc.model.getConfig().defaultFilmstripExpanded) {
                     onFilmstripToggle();
                 }
+                
+                // highlight and scroll to initially selected video
+                carousel.index = bcc.model.getConfig().playIndex;
```
tyard: I think you need to look at the playlist.getSelectedIndex() as opposed to the main model. That index will be set to whatever it should be (that's how the player is getting its video). Doing it this way means a) the two could potentially be out of sync and b) a default value will be ignored.

BC-Mikey: Fixed in new commit.

tyard:
looks good

BC-Mikey:
All merges made and ready to merge to dev

PR #91, opened 2015-07-24
CTA Fix: ford-127

BC-Mikey:


BC-Mikey:
Echoing my comments from JIRA 127

It appears that this is being caused by this line which hides the CTA Container everytime on modal load/video load:

https://github.com/BrightcovePS/FordPlayer/blob/dev/lib/bcgs_ford_modal.js#L218

I'm not exactly sure what the purpose of this conditional is. The comment specifies that it is to "show/hide" the 'bcc-has-ctas' class when the amount of visible CTAs is greater than 0 but the code only removes a class. Since this only gets called on load, this is basically saying "If no CTAs are visible (i.e., the details pane is defaulted to hidden), then we do not have CTAs"

I am going to remove this conditional.

Likewise, a similar CTA issue I also found which is very easy to fix and not worth creating another ticket is here:

https://github.com/BrightcovePS/FordPlayer/blob/dev/lib/bcgs_ford_modal.js#L213

This line is, again, only counting the amount visible upon modal load/video load. So, if this line is hit when the Details Pane is collapsed then we get more CTAs than specified in CTA_MAX.

lib/bcgs_ford_modal.js
```diff
@@ -210,14 +210,8 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
                 }(ctaConfig[i]));
                 
                 // Exits loop if the number of added CTAs is greater to the max allowed.
-                if ($ctaContainer.children().filter(":visible").length >= CTA_MAX) break;               
+                if ($ctaContainer.children().length >= CTA_MAX) break;
```
tyard: I don't think this is right for the requirements. CTAs can be added past the max allowed. The CSS controls which are visible -- sometimes there are some that are visible on larger screen sizes and some are only visible on smaller screen sizes. That's why this was set up to look at the amount visible as opposed to the number of children.

This might have broken when it changed to slideToggle() -- I found the disclaimer broke for a similar reason. visible was always returning false then, no matter what. So in that case I had to wait until the drawer was open before checking for visibility.

Long and short of it is that I don't think this change can happen as is, but I don't think the way it was set up before will work either as it was set up, especially now with slideToggle() being used.

BC-Mikey: I'm not sure I understand the purpose of CTA_MAX. From the code today it seems like CTA_MAX is there to prevent the details panel from showing more than what's configured. (In this case, CTA_MAX == 2) Whether they're visible or not, the plugin will only ever display up to 2 CTAs in based on this configuration. I didn't touch how this worked.

The issue I'm fixing here is that with using .filter("visible") here is that when a page loads with defaultDetailsExpanded == false, then filter("visible") does not equal true and the jQuery selector returns as 0 because none are visible. As a result, the next block is not run and the CSS class which should display the CTAs is never added:

```
if (!$ctaContainer.children().filter(":visible").length > 0) {
                $ctaContainer.removeClass("bcc-has-ctas");
}
```
If we want to keep the "visible" filter then it would make sense to run this code again when the details are toggled. As of now it only ever runs on MODEL_CHANGE, so when the details pane is hidden that first time, the first video is always recognized as not having any CTAs.

tyard: I think this needs to be reworked, yes. :visible used to work before the slideToggle() change. Now it returns false unless the drawer is opened (that's not how it used to work).

Desired end result: 

1) Only 2 CTAs should ever be visible at any time.
2) Which CTAs are visible depends on the device.

It is likely that the determination of CTA visibility can no longer be done via CSS, as was done originally, and it all needs to move to JS.

BC-Mikey: Ah, the slideToggle() function does add inline display styles to the .bcc-detail-bottom DIV, so that's probably why visible stopped working.

BC-Mikey: Opened revision #92 against this PR.

PR #93, opened 2015-07-27
FORD-143: reset filmstrip inline styles when modal is closed

BC-Mikey:
slideToggle() added inline styles which are hiding the filmstrip when it should be showing, however, the button is in the correct state. This has been removed by resetting the inline styles when the modal has been closed


tyard:
looks good

I might vote in the future to not use slideToggle() (just as I'd rather not use show/hide) because of the inline styles.

BC-Mikey:
I would agree with that assessment^

PR #102, opened 2015-08-04
Fix/closing share

bceacuna:
This PR contains the fix for FORD-153 and FORD-154. Note that I have removed the close button  from the Social Panel on smart phones since the close “X” button, in PSDs, is part of the lower panel, and not clickable


BC-Mikey:
This looks fine.

PR #103, opened 2015-08-05
centered social panel

bceacuna:
This PR contains the fixed for FORD-148 ([Mobile] iOS - Share widget is not centered)


tyard:
looks good

PR #104, opened 2015-08-06
localization for labels, spanish

tyard:


lib/bcgs_ford_main.js
```diff
@@ -47,6 +47,7 @@ v1.0.0: 6/2/2015
     config.modules = {
         Web: {},
         Util: {},
+        l10n: config.l10n,
         Slider: {},
```
bceacuna: Dont forget to add the bcc_l10n.js to the grunt build file

bceacuna:
Looks good. Just one comment.

lib/bcgs_ford_main.js
tyard: Right.

PR #106, opened 2015-08-06
fixed ford-164

bceacuna:
This PR contains the fix for FORD-164 (Android Tablets: Email window closes when click From/To entry fields)


tyard:
ok

PR #110, opened 2015-08-12
keyboard nav of button elements

tyard:


lib/bcgs_ford_filmstrip.js
```diff
@@ -250,7 +259,37 @@ v1.0.0: 6/19/2015
...
+                    break;
+                // ENTER, SPACE
+                case 13:
+                case 32:
+                    $activeDescendent.trigger("click");
+                    break;
+            }
+        });
+        
         return {
```
bceacuna: is there any reason why you don't use addKeyboardNav() to add all the attributes to the carousel?. 
// add role to the method definition first.
bcc.addKeyboardNav($carousel[0], "list", function(e){...}); //

lib/bcgs_ford_modal.js
```diff
@@ -437,20 +456,16 @@ TODO: remove instrinsic ratio implementation from index.html and it to the plugi
             
     $shareButton = $detailDiv.find('.bcc-share-icon');
     $saveButton = $detailDiv.find('.bcc-save-icon');         
+        $saveButton.on("click touchstart", invokeSaveFunction);
+        bcc.addKeyboardNav($saveButton[0], invokeSaveFunction);
+        $shareButton.on("click touchstart", openSocialPanel);
+        bcc.addKeyboardNav($shareButton[0], openSocialPanel);
```
bceacuna: indentation ?

lib/bcgs_ford_filmstrip.js
tyard: I'd have to add role (and to all invocations to this function) as well additional handlers for the left and right functionality (the function handles enter/space and would have to be done differently to allow for lists).

It's possible, but I think would actually make it more complex. I think if we get to the point of having multiple uses for keyboard nav for horizontal lists that we could add a new function for that, but for 99% of our use cases I think the current function is simpler.

lib/bcgs_ford_modal.js
tyard: That was intentional for readability. Breaks it up visually while keeping it as one block. I declared the vars on first lines, then added all the handlers to those buttons below.

lib/bcgs_ford_filmstrip.js
bceacuna: fair enough.

bceacuna:
Looks good. Merge away.

PR #111, opened 2015-08-12
reset addedIcons when player is launched

bceacuna:
This contains the fix for #FORD-144: Social Share icons are not always shown


tyard:
looks fine

PR #112, opened 2015-08-13
Email Panel fix

bceacuna:
This contains the fix for FORD-142: IE9/IE10/IE11- Email Share dialog need styling improvements

I made css selectors more specific so they are not overridden by the player


tyard:
all good

BC-Mikey:
Looks good!

PR #113, opened 2015-08-13
[iPad] removed Selection Area

bceacuna:
This contains the fix for FORD-176: Do not display characters on top of video player after a copy selection is made


tyard:
all good

PR #115, opened 2015-08-18
Feature/text navigation

bceacuna:
Implemented 
https://bcconsulting.atlassian.net/browse/FORD-179: Enable keyboard navigation for text elements


lib/bcc_web.js
```diff
@@ -199,6 +211,7 @@ v1.1.0: 8/12/2015
         toggleFullscreen: toggleFullscreen,
         isFullscreen: isFullscreen,
         addKeyboardNav: addKeyboardNav,
+        addWAIARIA: addWAIARIA,
```
tyard: I think we need a friendlier name for this. addWAIARIA, even if that is what it is, is really hard to scan and parse. But tabindex isn't aria anyway. How about making this one the addKeyboardNav (because it contains the tabindex) and making the original addKeyboardNav into something like addKeyboardNavForButton?

tyard: Other than that it looks good.

bceacuna: Make sense.

